PR #66, opened 2023-04-30
Trivial fix

dyr429:
Add training component controller in similar pattern as others

Add HTML training to MVNV demo
image

Add image training to CM demo
image

Add reset button in case of routing error
image

Plus some trivial fix.


public/configs/config-cleveland.hjson
```diff
@@ -182,38 +186,38 @@
...
+//                }
+////text example
+//                {
+//                    id: "cm-response4"
+//                    type: "shortText"
+//                    prompt: "Your answer!!"
+//                    desc: "Enter your short text answer here"
+//                    required: true
+//                    location: aboveStimulus
+//                }
```
JackWilb, line 220: Let's remove this before merge

JackWilb (approved):

public/configs/config-cleveland.hjson
dyr429: Done. Change questions to real survey questions.

PR #78, opened 2023-05-04
Set iframe height when the stimuli is ready

### LINKED ISSUE #72: Fix iframe height

hilsonshrestha:
Closes #72 

Receives an event with document height when when iframe stimuli is ready.
Iframe height is then set by the IframeController.

image

### DIFF
--- a/src/controllers/IframeController.tsx
+++ b/src/controllers/IframeController.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useMemo } from 'react';
+import { useEffect, useMemo, useRef } from 'react';
 import { useDispatch } from 'react-redux';
 import { useParams } from 'react-router-dom';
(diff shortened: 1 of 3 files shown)


src/controllers/IframeController.tsx
```diff
@@ -58,8 +59,10 @@ const IframeController = ({
               type: trialConfig?.type,
             })
           );
-
-          // navigate(`/${computedTo}`, { replace: false });
+        } else if (data.type === `${PREFIX}/READY`) {
+          if (ref.current) {
+            ref.current.style.height = `${data.message.documentHeight}px`;
+          }
```
JackWilb, line 65: Is this a data race? Would this ever not work as we expect?

JackWilb (approved):

PR #102, opened 2023-05-05
Update image for cmg training

### LINKED ISSUE #101: CMG training looks odd
Image

It's grainy; we need this to be higher quality. 

It doesn't use the right UI.

hilsonshrestha:
Closes #101 

image


alexsb (approved):
Looks great; should we also check in the source material?

PR #104, opened 2023-05-05
Move cmg input to sidebar

### LINKED ISSUE #99: In CMG, make practice consistent with trials in terms of layout
This is practice: 

Image

This is Trials: 

Image

They should look identical, except for the additional UI elements.

hilsonshrestha:
Closes #99 

TODOs:
- [ ] Check answer and Input button is disabled. This needs to be enabled.

### DIFF
--- a/public/configs/config-cleveland.hjson
+++ b/public/configs/config-cleveland.hjson
@@ -50,21 +50,21 @@
                 interval: 10
             }
             // nextButtonLocation: 'sidebar' | 'aboveStimulus' | 'belowStimulus'
-            nextButtonLocation: belowStimulus
-            instructionLocation: aboveStimulus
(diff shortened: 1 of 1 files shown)


alexsb (approved):

src/components/response/ResponseBlock.tsx
```diff
@@ -53,7 +53,7 @@ export default function ResponseBlock({
 
   const responses = config.response.filter(
     (r) =>
-      (!r.location && location === 'belowStimulus') || r.location === location
```
JackWilb, line 56: Since location is optional, the default was 'belowStimulus'. This line change breaks that behavior and if a response is declared without location, it will no longer render

Issue #90, opened 2023-05-04
Change public folder structure from data-type specific folders, to experiment specific


alexsb:
This doesn't make sense:

![Image](https://user-images.githubusercontent.com/2039375/236351035-7be91734-20c8-4bc0-8848-dac02b07a9e5.png)

There should be one folder, typically with the same name as the config, with all the content in it. 

Chan

PR #106, opened 2023-05-05
Add figma ref

hilsonshrestha:


alexsb (approved):
OK for now, but we should move this to a different folder when #90 is done.

PR #109, opened 2023-05-05
adding sync to firebase feature

### LINKED ISSUE #91: Improve end-page
This should be configurable form the config, but by default say: 

"Thank you for completing the study. You may close this window now." Centered horizontally and vertically.

kirangadhave:
Closes #91 

syncs the entire trrack provenance to Firebase
load an existing session if found
ask the user if they want to load or start a new session

Future:
- autoload session if less than X minutes (@alexsb thoughts?)
- accept firebase details from spec


src/components/ErrorBoundary.tsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+
+/**
+ * Helper component for debugging. Should not be used always
+ */
+export default class ErrorBoundary extends React.Component<any, any> {
```
dyr429, line 6: Does this have to be a class component?

dyr429 (approved):
Before merging,  you can comment out debugging console log.

src/components/ErrorBoundary.tsx
kirangadhave: Yeah it has to be. It uses some of the lifestyle features which are only available in class components.

Issue #114, opened 2023-05-06
Survey study is completely broken


alexsb:
Just shows an empty page.

alexsb:
This still shows nothing on the deployed page: 



![Image](https://github.com/revisit-studies/study/assets/2039375/fe17a219-0e9a-482d-b2f2-75c4ef7076e2)

JackWilb:
I see this is working on the deployed version
<img width="1624" alt="Screenshot 2023-05-15 at 2 22 29 PM" src="https://github.com/revisit-studies/study/assets/36867477/c4010b32-818d-4360-8c98-19081117cbe2">

Issue #95, opened 2023-05-05
Use "description" and title field in admin panel.


alexsb:
Use title and description instead of "instruction". Instructions can be too long. 

![Image](https://user-images.githubusercontent.com/2039375/236355543-9ed3e7a4-ade0-4047-a115-4686b4947b4f.png)

PR #125, opened 2023-05-09
Config fix

### LINKED ISSUE #103: Add title and metadata to trial object
Thinking about how we can generate good data from our experiments, I think we need a few more field in the trial object. Here's what we have right now.  

```
 "dotplot-low": {
          // The description is used in the admin panel and not shown to particpants. 
          "description": "Task 1: low temperatures.",
          "instruction": '''
```

Here's what we should have: 

```
 "dotplot-low": {
          // The description is used in the admin panel and not shown to particpants. 
          "title": "T1-low-temp"
          "properties": [
               temp: low,
               difficulty: 1,
               nr-dots: 18
          }
          "description": "Task 1: low temperatures.",
          "instruction": '''
```

The content of properties is a flat array of key-value pairs. These will then be mapped to observations in tidy data.

dyr429:
Closes #103 #114 #95

- Fix bugs on survey demo
- Add meta to our demos, now meta data can be displayed in admin panel along with  ID and description

imageimage

### DIFF
--- a/src/parser/types.ts
+++ b/src/parser/types.ts
@@ -63,6 +63,7 @@ export interface SurveyComponent extends StudyComponent {
 }
 
 export interface Trial {
+  meta?: object;
   description: string;
   instruction: string;
   stimulus: Stimulus;
--- a/src/parser/types.ts
+++ b/src/parser/types.ts
@@ -63,7 +63,7 @@ export interface SurveyComponent extends StudyComponent {
 }
 
 export interface Trial {
-  meta?: object;
+  meta?: Record<string,any>;
   description: string;
   instruction: string;
   stimulus: Stimulus;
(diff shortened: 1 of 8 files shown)


src/parser/types.ts
```diff
@@ -63,6 +63,7 @@ export interface SurveyComponent extends StudyComponent {
 }
 
 export interface Trial {
+  meta?: object;
```
kirangadhave, line 66: Can you change this to Record<string, any> type? 

Using object as type leads to issues. A record expects the key will be string, but object requires named keys, so it only allows using inbuilt keys.

kirangadhave (changes requested):

kirangadhave (approved):

PR #133, opened 2023-05-15
Clean global config

### LINKED ISSUE #110: Clean out Global Config
The global config shouldn't contain anything but the path to the other config file; all of the data should be loaded from there.

hilsonshrestha:
Closes #110 

Global config will now only have paths to study configs. All study configs are moved within their respective folders.


public/html-demo/config-html-demo.hjson
```diff
@@ -1,6 +1,7 @@
 {
     configVersion: 1
     studyMetadata: {
+        urlKey: "html-demo"
```
JackWilb, line 4: This isn't required, right?

JackWilb (approved):

PR #148, opened 2023-06-06
Help-content-update

dyr429:
- Update help content in help modal.
- Add image support for the help modal, mvnv study uses the original help image.
- Also fix the iframe input issue.
- 
image
image


public/mvnv/config-html-MVNV.hjson
```diff
@@ -17,7 +17,8 @@
     }
     uiConfig: {
         contactEmail: "test@test.com"
-        helpTextPath: "mvnv/help.md"
+        helpTextPath: "mvnv/help.md",
+        helpImgPath: "mvnv/quickStart.png",
```
JackWilb, line 21: Can you not reference this image from the help.md? If we could do that, it would mean it would be possible to embed many images, instead of being limited to 1

JackWilb (commented):
It looks good so far, but I think we can refactor how we handle images in the help file. We should just embed the images using markdown embedding, instead of adding a custom image attribute to the config.

JackWilb (approved):

PR #152, opened 2023-06-16
Vlat reactinput demo

dyr429:
Add Two demo studies

1. VLAT using image stimuli
2. Click accuracy test, using react component stimuli. An example of passing answers from react component to Revisit.

image

image


src/components/stimuli/ClickAccuracyTest.tsx
```diff
@@ -0,0 +1,98 @@
+import * as d3 from 'd3';
+import { Bubbles } from './chartcomponents/Bubbles';
+import { DotMarks } from './chartcomponents/DotMarks';
```
JackWilb, line 3: Unused imports Bubbles and DotMarks

JackWilb (approved):

PR #161, opened 2023-08-31
Allowing for validation in consent form

ZachCutler04:
Closes #

Allows for validation of responses before continuing, such as for the consent form to confirm someone accepted. 

Currently the types which check this in the config, requiredValue and requiredLabel, are on the radio button type. Ideally this would be generalized to all response types, but seeing as value can be many different things (an array, number, etc), I left it on just radio for now. 

https://github.com/revisit-studies/study/assets/24012760/5f04df9c-947c-4452-aac8-dc63569269b9

TODOs:


public/cleveland/config-cleveland.hjson
```diff
@@ -45,6 +45,8 @@
                     id: accept
                     prompt: Do you accept the consent form?
                     required: true
+                    requiredValue: yes
+                    requiredLabel: Accept
```
JackWilb, line 49: Maybe we could just do this with value?

src/components/response/ResponseBlock.tsx
```diff
@@ -140,9 +141,8 @@ export default function ResponseBlock({
   return (
     <div style={style}>
       {responses.map((response) => (
-        <>
+        <React.Fragment key={`${response.id}-${id}`}>
```
JackWilb, line 144: Does this stop some re-rendering?

src/parser/schema.json
```diff
@@ -877,6 +904,7 @@
     },
     "WebsiteComponent": {
       "additionalProperties": false,
+      "description": "The WebsiteComponent interface is used to define the properties of a website component. A WebsiteComponent is used to render an iframe with a website inside of it. This can be used to display an external website or an html file that is located in the public folder.",
```
JackWilb, line 907: Are these programmatically generated?

JackWilb (changes requested):

src/parser/schema.json
ZachCutler04: Ya I was also wondering why something changed here? I thought maybe i installed a updated version of whatever library did this or something?

src/components/response/ResponseBlock.tsx
ZachCutler04: Theoretically yes, it could, but really it's just to get rid of the console error which was saying theres duplicate keys in a list, since the fragment is technically the part of the list which the key is actually checked on since its the top element

ZachCutler04:
Made a few changes, first I moved the requiredValue and requiredLabel to the BaseResponse, so that every type can be given a requiredValue. Currently the requiredValue works on every type that I tried, including the checkbox (which is an array). 

We discussed removing the requiredLabel, but as many of the response types dont have set labels, I left it in as optional. For example, a Likert response doesnt have a set label, but could have a requiredValue, thus the need for the requiredLabel. Obviously there's no reason for this to ever actually happen, but still. I did change the logic to look for a label in the options array for RadioButtons, but thats specific to radio buttons. 

I also couldnt actually find an example of a checkbox, so added one to the config-survey-demo.

src/components/response/RadioInput.tsx
```diff
@@ -22,8 +22,10 @@ export default function RadioInput({
         withAsterisk={required}
         size={'md'}
         {...answer}
+        // This overrides the answers error. Which..is bad?
+        error={answer.value && requiredValue && requiredValue.toString() !== answer.value.toString() ?  `Please select ${requiredLabel ? requiredLabel : options.find((opt) => opt.value === requiredValue)?.label} to continue.` : null}
```
JackWilb, line 26: We talked about moving this off the radio input and onto the ResponseBlock or ResponseSwitcher. Is that handled by the utils? If so, why is this here?

JackWilb: Also, does `&& requiredValue.toString() !== answer.value.toString() ? Please select ${requiredLabel ? requiredLabel : options.find((opt) => opt.value === requiredValue)?.label} to continue. : null` fail to validate if there is no `requiredValue`?

JackWilb: And a third thing 🙃 How does this validation work with arrays of required values?

src/components/stimuli/inputcomponents/utils.ts
```diff
@@ -19,8 +19,16 @@ const generateValidation = (responses: Response[], id: string) => {
         ...validateObj,
         [`${id}/${response.id}`]: (value: string | string[]) => {
           if (Array.isArray(value)) {
+            if(response.requiredValue && !Array.isArray(response.requiredValue)) {
+              return 'Incorrect required value';
+            } else if (response.requiredValue && Array.isArray(response.requiredValue)) {
+              return response.requiredValue.length === value.length && response.requiredValue.every((val, index) => val === value[index]) ? null : 'Incorrect input';
```
JackWilb, line 25: This requires that the arrays have the same order. Is that guaranteed?

src/components/response/RadioInput.tsx
ZachCutler04: This part doesn't really validate, the actual validation is handled entirely in the utils. This is just the UI part that shows the error message in the RadioInput itself. Which is why it doesnt do anything if there is no requiredValue, and why I dont worry about arrays.

src/components/stimuli/inputcomponents/utils.ts
ZachCutler04: True, changed to sort them first (in a new array, not sure if I can safely sort the response/value array in place?)

src/components/response/RadioInput.tsx
JackWilb: Gotcha. Should we do this on the other input types too, then?

JackWilb (approved):
Approving with 1 note. Perhaps we should make that an issue so we can get this in and you can iterate. The tests will also need to be fixed up

PR #177, opened 2023-10-31
Task provenance

ZachCutler04:

Adds the ability to create a trrack graph and events for individual trials


src/components/stimuli/ClickAccuracyTest.tsx
```diff
@@ -14,15 +16,45 @@ const chartSettings = {
...
+interface ClickAccuracyTest {
+  distance: number;
+  speed: number;
+  clickX: number;
+  clickY: number;
+}
+
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-const ClickAccuracyTest = ({ parameters, trialId }: { parameters: any, trialId: string }) => {
+const ClickAccuracyTest = ({ parameters, trialId, setAnswer }: { parameters: any, trialId: string, setAnswer: ({trialId, status, provenanceGraph, answers} : {trialId: string, status: boolean, provenanceGraph?: ProvenanceGraph<any, any, any>, answers: Record<string, any>}) => void }) => {
```
JackWilb, line 27: Can we make this type look nicer? Maybe extract the type to a locally declared type above line 27?

src/components/stimuli/ClickAccuracyTest.tsx
```diff
@@ -14,15 +16,45 @@ const chartSettings = {
...
+        registry: reg,
+        actions: {
+          click
+        },
+    };
+}, []);
+
+  const trrack = useMemo(() => {
+    return initializeTrrack({registry: registry, initialState: {distance: 0, speed: 10, clickX: 0, clickY: 0} });
+  }, [registry]);
```
JackWilb, line 56: `useMemo` seems unnecessary here, right? Or maybe I just don't understand react well enough haha 🤷🏻

src/components/stimuli/ClickAccuracyTest.tsx
JackWilb: This could also be part of the API that we pass down to these stimulus controllers to help our users integrate with the platform

src/controllers/ReactComponentController.tsx
```diff
@@ -25,6 +26,7 @@ const ReactComponentController = ({
       <StimulusComponent
         parameters={parameters}
         trialId={trialId}
+        setAnswer={setAnswer}
```
JackWilb, line 29: This is the location of the API for the users' stimulus components. Nothing todo here, just making a note for us

src/store/flags.ts
```diff
@@ -63,12 +66,17 @@ const flags = createSlice({
...
           },
           answers: {},
+          provenanceGraph: null,
         };
       }
       state.trialRecord[payload.trialId].valid[payload.location] =
         payload.status;
       const prev = state.trialRecord[payload.trialId].answers;
 
+      if(payload.provenanceGraph) {
```
JackWilb, line 76: I prefer these if checks to be explicit

```suggestion
      if(payload.provenanceGraph !== null) {
```

src/store/flags.ts
```diff
@@ -105,6 +113,16 @@ export const useFlagsSelector: TypedUseSelectorHook<FlagStore> =
 export const useFlagsDispatch: () => FlagDispatch =
   createDispatchHook<FlagStore>(flagsContext);
 
+export function setAnswer({trialId, status, provenanceGraph, answers} : {trialId: string, status: boolean, provenanceGraph?: ProvenanceGraph<any, any, any>, answers: Record<string, any>}) {
```
JackWilb, line 116: Here is that same ugly type. Let's move this into types.ts and import it in both places

JackWilb (changes requested):
Some clean up required, but overall looks pretty simple. Thanks, Zach!

src/store/flags.ts
ZachCutler04: Fixed up types

src/store/flags.ts
ZachCutler04: done

src/components/stimuli/ClickAccuracyTest.tsx
ZachCutler04: You definitely do want a memo here I think, otherwise you'd initialize trrack on every render. But you also definitely don't need two idk why i did it like this, cleaned it up to do this in the same memo

src/components/stimuli/ClickAccuracyTest.tsx
ZachCutler04: So I moved the types of the entire parameters for these stimulus components into the types, typed the clickAccuracy as StimulusParameters, and did some type magic in the flags to make that work. So now we actually have some intelisense in the Stimulus components, which is cool!

JackWilb (approved):

PR #211, opened 2023-11-08
Merge trial component with base component config before tidy download

hilsonshrestha:
Merge each component's config with base component's config before tidy download.


ZachCutler04 (approved):
Looks good

PR #218, opened 2023-11-08
Add Fairness JND study

hilsonshrestha:


ZachCutler04 (approved):
looks good, no changes to core code

PR #221, opened 2023-12-04
Decouple next stimuli logic

hilsonshrestha:
Decouple the logic behind storing current stimuli responses and proceeding to next stimuli.
This makes it feasible for any stimili react components to create their own buttons / key press events to proceed to next stimuli. 

TODOs:
- [ ] Test with iframes and database store


src/controllers/IframeController.tsx
```diff
@@ -33,7 +33,7 @@ export default function IframeController({ currentConfig }: { currentConfig: Web
   // navigation
   const currentStep = useCurrentStep();
   const navigate = useNavigate();
-  const computedTo = useNextStep();
+  const { nextStep: computedTo } = useNextStep();
```
JackWilb, line 36: It will be cleaner and easier to debug this code if we just change the variable to `nextStep`]

```suggestion
  const { nextStep } = useNextStep();
```

src/components/response/ResponseBlock.tsx
```diff
@@ -31,29 +31,23 @@ export default function ResponseBlock({
...
   const hasCorrectAnswer = ((configInUse?.correctAnswer?.length || 0) > 0);
 
-  const startTime = useMemo(() => {
-    return Date.now();
-  }, []);
-
   const showNextBtn =
     location === (configInUse?.nextButtonLocation || 'belowStimulus');
 
+  // TODO: shift following effects to useNextStep hook, fix dependencies
```
JackWilb, line 50: TODO?

src/store/hooks/useNextStep.ts
```diff
@@ -1,17 +1,105 @@
...
+    startTime,
+    currentStep,
+    saveTrialAnswer,
+    computedTo,
+  ]);
 
-  return nextStep || 'end';
+  return {
+    nextStep,
+    isDisabled: isNextDisabled,
```
JackWilb, line 102: I prefer to keep things with the same name for easier debugging later (1 global search instead of 2). Let's change this to:
```suggestion
    isNextDisabled,
```
And then update it everywhere else

src/store/hooks/useNextStep.ts
```diff
@@ -1,17 +1,105 @@
...
+} from '../store';
+import { useCurrentStep, useStudyId } from '../../routes';
+import { useCallback, useMemo } from 'react';
+
+import { useNavigate } from 'react-router-dom';
+import { deepCopy } from '../../utils/deepCopy';
+import { ValidationStatus } from '../types';
+import { useStorageEngine } from '../storageEngineHooks';
+import { useStoredAnswer } from './useStoredAnswer';
+// import { useStoreSelector } from '../store';
```
JackWilb, line 15: Remove this commented line

src/components/response/ResponseBlock.tsx
```diff
@@ -150,14 +103,9 @@ export default function ResponseBlock({
...
         {showNextBtn && (
           <NextButton
-            disabled={
-              hasCorrectAnswer
-                ? !checkClicked
-                : !areResponsesValid
-            }
-            process={() => {
+            disabled={hasCorrectAnswer && !checkClicked}
+            onClick={() => {
```
JackWilb, line 107: There is an onclick here and in the button itself, is that an issue? Is the order that they execute in fixed? Does that matter?

src/components/response/ResponseBlock.tsx
```diff
@@ -150,14 +103,9 @@ export default function ResponseBlock({
         )}
         {showNextBtn && (
           <NextButton
-            disabled={
-              hasCorrectAnswer
-                ? !checkClicked
-                : !areResponsesValid
-            }
-            process={() => {
+            disabled={hasCorrectAnswer && !checkClicked}
```
JackWilb, line 106: Maybe we could us another name instead of `disabled`. Perhaps `correctAnswerChecked` would be a better name

src/controllers/IframeController.tsx
JackWilb: Actually, I'm not sure if we even need this. It's just used as a dependency of the useEffect below, but since we also have currentStep, it should be okay without it

JackWilb (changes requested):

src/components/response/ResponseBlock.tsx
hilsonshrestha: Yeah, I am unsure if this should be shifted to useNextStep hook as well or not.

src/components/response/ResponseBlock.tsx
hilsonshrestha: I haven't tested it, but order might be important here.
`onClick` might be dependent on the a ui component. And if it is called after the next screen has been loaded (previous screen is unmounted), we might get an error.

src/components/response/ResponseBlock.tsx
hilsonshrestha: @JackWilb , I will make a separate PR for this. I think all of the effects should be shifted to useNextStep hook, which will also return user's answers and correct answer if available.

This makes it easier feasible to completely remove ResponseBlock if study designer wants to create their own within the stimuli.

PR #232, opened 2023-12-20
Track all events for later rehydration

JackWilb:
No

This tracks all mouse, keypresses, scroll, change tab, etc. events so that they can be rehydrated later. The interval can be modified in the `studyConfig.uiConfig` using `windowEventDebounceTime` which takes a time in ms to debounce by. For example, `100` would give events 10 times per second and `200` would be 5 times per second.

The way that the windowEvents are saved is similar to the provenance. That is, stored as an array on the answers for a particular step, so it can be retrieved in a similar way. Additionally, this makes the logic to save and load the events the same as the provenanceGraph. As such, I've simplified the logic for these two properties in the firebaseStorageEngine to remove as much code duplication as possible

TODOs:
- [x] Update relevant documentation (#233)
- [x] Add some options for these events to the config
- [x] Do something with the data, save it somewhere
- [x] Add storageEngine method to get all event data
- [x] Consolidate the firebase storage getters and setters into 1 function

### DIFF
--- a/public/html-demo/config-html-demo.hjson
+++ b/public/html-demo/config-html-demo.hjson
@@ -21,6 +21,7 @@
         withProgressBar: true
         autoDownloadStudy: false
         sidebar: true
+        windowEventDebounceTime: 200
     }
     components: {
         introduction: {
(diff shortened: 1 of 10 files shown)


ZachCutler04 (approved):
Code looks good, tested with brush interaction and saving with firebase works great.

PR #251, opened 2024-02-20
Fix sequence array size limitations and data races

### LINKED ISSUE #176: Fix data race problems in the randomization logic
There's a few different data race issues in the randomization logic that could cause small separation from the expected latin squares. Should address those.

### LINKED ISSUE #250: Large sequence order may exceed max size allowed in firebase
**Describe the bug**
With large number of stimuli and randomization, the sequence array may exceed max allowed size in the firebase

**Screenshots**
image

JackWilb:
Closes #176 
Closes #250 

This fixes the issues with the sequence array:
1. The sequence array could be too large. I fixed this by using storage, which has no real size limitation.
2. Race conditions when pulling a sequence. I fixed this in 2 ways. The first fix is to not modify the sequence array, thus we can just get an index and take the sequence at that index as our sequence. Second, I made the firebase database track a timestamp of a participants intent to get a sequence, which can then be used to find the order in which people state their intent. We can use these 2 fixes together to avoid any race conditions.

N/A

TODOs:
- [x] Update relevant documentation (number of sequences generated)
- [x] Make the number of sequences generated modifiable through the config
- [x] What do we do if the intent array is messy? That might ruin some of the guarantees (#255)

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -23,7 +23,7 @@ jobs:
             cache_dir: ~/.cache/ms-playwright
 
           - project: webkit
-            os: macos-latest
+            os: macos-14
             cache_dir: ~/Library/Caches/ms-playwright
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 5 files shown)


ZachCutler04 (approved):
Looks good

PR #254, opened 2024-02-20
Add skip logic to the configuration

### LINKED ISSUE #182: Add conditional branching
Add branching of sequence based on responses of the users

### LINKED ISSUE #234: Do we add `name` field for each task in the config?
Currently the config does not have a field `name`, and we are using `key` for the sidebar.
Should we instead have a separate field like `name` in the config for each stimuli?

image

JackWilb:
Depends #256 
Closes #182
Closes #234 

This PR adds skip logic to the configuration using a new key (`skip`) in the ComponentBlocks of the order. The skip conditions work as you would expect and are described in detail in the src/parser/types file. There are 4 types of checks that I've enabled with this PR:

1) IndividualComponentSingleResponseCondition: This checks one response on a component for a specific value.
2) IndividualComponentAllResponsesCondition: This checks one component to make sure all answers are correct (as defined in the correctAnswer in the study config)
3) ComponentBlockCondition: This checks all components within a block to make sure all answers are correct (as defined in the correctAnswer in the study config). This will then skip based a number of correct or incorrect responses.
4) RepeatedComponentBlockCondition: This checks all components with the same name within a block to make sure all answers are correct (as defined in the correctAnswer in the study config). Again, this skips based a number of correct or incorrect responses across these components. This is most useful for attention checks.

It's possible to specify multiple of these conditions on a block and the code handles that by checking all of the conditions and skipping when one matches. Additionally, the code handles nested block well and can trigger the parent skip block to fire if it is satisfied by answers in the child block. 

I've added a rigorous test for this addition, which you should run (with the browser open) to watch how the skip happens. Since the skip is implemented in the useNextStep hook, it calculates right as the participant clicks the next button.

Finally, I took a pass to clean up the code so hopefully this PR is easy to review.

N/A

TODOs:
- [x] Update relevant documentation


JackWilb:
Consider attention checks

src/utils/getSequenceFlatMap.ts
```diff
@@ -1,6 +1,68 @@
...
+  firstIndex: number;
+  lastIndex: number;
+};
+
+function _findBlockForStep(sequence: Sequence, step: number, distance: number): (SkipConditionWithExtents[]) | number {
+  let componentsSeen = 0;
+  for (let i = 0; i < sequence.components.length; i += 1) {
+    const component = sequence.components[i];
+    if (typeof component === 'string') {
+      if (step === distance + componentsSeen) {
```
bbollen23, line 19: Not sure how you feel about this -- I always do this because it makes it more readable for me. Up to you.
```suggestion
      if (step === (distance + componentsSeen) ) {
```

src/store/hooks/useNextStep.ts
```diff
@@ -88,24 +111,89 @@ export function useNextStep() {
...
+    // Get the answers with the new answer added, since above is dispatching and async, but we need it synchronously
+    const answersWithNewAnswer = {
+      ...answers,
+      [identifier]: {
+        answer, startTime, endTime, provenanceGraph, windowEvents: currentWindowEvents,
+      },
+    };
+
+    // Check if the skip block should be triggered
+    if (hasSkipBlock && answersWithNewAnswer[identifier]) {
```
bbollen23, line 131: Is `answersWithNewAnswer[identifier]` necessary here? It looks like above, this must always be defined.

bbollen23 (changes requested):
I'll be working on checking this branch out locally and testing so I can get a better feel for the changes. Otherwise, code looks prime

src/store/hooks/useNextStep.ts
JackWilb: Fixed in 62e610f

bbollen23 (approved):

ZachCutler04 (approved):
Reviewed in person with Jack, looks great

PR #273, opened 2024-03-21
Update example of data passing between reVISit and HTML

hilsonshrestha:
-

There was an example of passing data from HTML to reVISit, but there was no example of how to get the data (config parameters) from revisit to the html files.

-

-

### DIFF
--- a/public/html-input/bar-chart-interaction.html
+++ b/public/html-input/bar-chart-interaction.html
@@ -14,7 +14,8 @@
 
   <script>
     // Get data from the config file
-    Revisit.onDataReceive(function ({ data }) {
+    Revisit.onDataReceive((data) => {
+      const barData = data['barData']
       const taskID = 'barChart';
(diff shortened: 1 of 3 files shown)


bbollen23:
@JackWilb @hilsonshrestha Adjusted this example to align with the tutorial we will add to the reVISit site. Note that the example that we will have in the tutorial will build off of a previously done tutorial (a simple "questionnaire"). In other words, the tutorial is written so that a user first creates a simple questionnaire study, then they will add these bar charts in the next tutorial to the already existing tutorial -- not creating a whole new one. Because of this, the only way for these to exactly match is if we changed this entire example to also have the questionnaire bits. I've instead elected to just adjust the HTML to be identical to what the tutorial will have for the HTML portion and adjusting the configuration file to be written in the same way. @JackWilb Let me know if this is what you were thinking.

PR #287, opened 2024-04-02
Participant data docs

### LINKED ISSUE #281: Add ParticipantData to the typedoc generation
To better document the fields that are available in the JSON export, we can add ParticipantData to the typedoc. We would then link to this from the tutorial page to simplify the data export section

bbollen23:
Adds ParticipantData documentation and the documentation for relevant sub types. 
Closes #281 

For this, decided to export the ParticipantDat, StoredAnswer, and Sequence interfaces in the parser/types.ts file so we could still have just one entry point for the tyepdoc.json file. Added the relevant eslint commands to disable errors on the circular imports that were caused.

### DIFF
--- a/src/storage/types.ts
+++ b/src/storage/types.ts
@@ -1,3 +1,4 @@
+// eslint-disable-next-line import/no-cycle
 import { Sequence, StoredAnswer } from '../store/types';
 
 /**
--- a/src/storage/types.ts
+++ b/src/storage/types.ts
@@ -33,12 +33,17 @@ Each key in answer will be labeled the same as the response component that it re
         },
         "startTime": 1711641174858,
         "endTime": 1711641178836,
+        "provenanceGraph":{
+          ...
+        },
         "windowEvents": [
           ...
         ]
       }
 ```
 The keys of this object are the names of the components with an additional underscore and number appended to the end. This is done so that the study creator can discern between not only the components but also between the various instances of the same component when necessary. All times are in **epoch milliseconds**.
+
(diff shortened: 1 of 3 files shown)


src/storage/types.ts
```diff
@@ -1,9 +1,56 @@
...
+        "answers": {
+            ...
+        },
+        "searchParameters": {
+            ...
+        }
+    }
+]
+```
+Each key in answer will be labeled the same as the response component that it refers to. The sequence shows the order that the participant saw each component (since these may be different for each participant if the configuration sequence has some randomization). This answer wil contain information such as the start time, the end time, and all of the window events. See the example below.
```
JackWilb, line 25:
```suggestion
Each key in answer will be labeled the same as the response component that it refers to. The sequence shows the order that the participant saw each component (since these may be different for each participant if the configuration sequence has some randomization). This answer will contain information such as the start time, the end time, and all of the window events. See the example below.
```

src/store/types.ts
```diff
@@ -27,11 +28,65 @@ export type TrialValidation = Record<
...
+Each item in the window event is given a time, a position an event name, and some extra information for the event (for mouse events, this is the location).
+*/
 export interface StoredAnswer {
+  /** Object whose keys are the "id"s in the Response list of the component in the StudyConfiguration and whose value is the inputted value from the participant. */
   answer: Record<string, Record<string, unknown>>;
+  /** Time that the user began interacting with the component in epoch milliseconds. */
   startTime: number;
+  /** Time that the user ended interaction with the component in epoch milliseconds. */
   endTime: number;
   provenanceGraph?: TrrackedProvenance,
```
JackWilb, line 59: We should document `provenanceGraph` here and make a note of it in the ParticipantData. You can say that is related to the trrack provenance graph that is saved from the react components. We don't need to go into too much detail, but should at least mention it

JackWilb (changes requested):

JackWilb (approved):

Issue #291, opened 2024-04-05
Determine how to disable routing on studies


JackWilb:
Don't let a user navigate except through the next button

bbollen23:
Closed since we've add the "disable browser back".

bbollen23:
Re-opening. May use other methods to alert user when navigating through URL or tracking such info.

PR #293, opened 2024-04-09
Google auth

### LINKED ISSUE #290: Add admin user with gated access to admin panel and admin mode

### LINKED ISSUE #292: Add Google SSO using FIrebase for Firebase Storage Engine Users
Need an authentication and authorization flow so that study creators can separate admins from anon users.

bbollen23:
Closes #290 
Closes #292 

This PR sets up a Google SSO Authentication flow for users who are using Firebase as their storage engine. The analysis dashboard and the main page (displaying all the studies) are not behind protected routes which require login. This functionality is disabled whenever the storage engine is NOT firebase. The user provides a list of admins in the global.json in the key "adminUsers". These entries must be valid Google emails. The user info is stored in local storage so a login will persist through refreshes and restarting the server. The individual studies are not behind protected routes as they are meant to be used by completely anonymous users. Below we can see some simple functionality, including refreshing does not interfere with the user session, error codes for non-admin users (and other error codes such as a change in the debug token) are displayed on the login page properly. 

https://github.com/revisit-studies/study/assets/26078758/18e0024e-c6f3-4e12-9423-9260fca59cd1
https://github.com/revisit-studies/study/assets/26078758/21872986-8dae-408c-8008-3591e84736d2

TODOs:
- [ ] #291 
- [ ] Develop and implement workflow for storing admin users in the Firebase store rather than in the global.json (best workflow might be adding the admin users to the JSON and letting the application store them directly in Firebase and then reading from Firebase in subsequent sessions).

### DIFF
--- a/public/global.json
+++ b/public/global.json
@@ -51,5 +51,8 @@
         "demo-image": {
             "path": "demo-image/config.json"
         }
-    }
+    },
+    "adminUsers":[
+        "briancbollen@gmail.com"
+    ]
 }
--- a/src/ProtectedRoute.tsx
+++ b/src/ProtectedRoute.tsx
@@ -1,17 +1,30 @@
-import { useState, useEffect } from 'react';
-import { Navigate, Outlet } from 'react-router-dom';
-
-export interface User {
-  email: string;
-  admin: boolean;
-  name?: string;
-}
+import { ReactNode, useEffect } from 'react';
+import { Navigate } from 'react-router-dom';
+import { useAuth, User } from './store/hooks/useAuth';
+import { useStorageEngine } from './store/storageEngineHooks';
 
 interface ProtectedRouteProps {
-  user: User | undefined;
-  isAuth: boolean;
+  children: ReactNode;
 }
 
-export function ProtectedRoute({ user, isAuth }: ProtectedRouteProps) {
-  return isAuth ? <Outlet /> : <Navigate to="/login" />;
+export function ProtectedRoute({ children }: ProtectedRouteProps) {
+  const { user, login } = useAuth();
+  const { storageEngine } = useStorageEngine();
+
+  useEffect(() => {
+    if (storageEngine?.getEngine() !== 'firebase') {
+      // If not using firebase as storage engine, authenticate the user with a fake user with admin privileges.
+      const currUser: User = {
+        name: 'localName',
+        admin: true,
+        email: 'localEmail@example.com',
+      };
+      login(currUser);
+    }
+  }, []);
+
+  if (user && user.admin) {
+    return <div>{children}</div>;
+  }
+  return <Navigate to="/login" />;
 }
--- a/src/ProtectedRoute.tsx
+++ b/src/ProtectedRoute.tsx
@@ -1,27 +1,13 @@
-import { ReactNode, useEffect } from 'react';
+import { ReactNode } from 'react';
 import { Navigate } from 'react-router-dom';
-import { useAuth, User } from './store/hooks/useAuth';
-import { useStorageEngine } from './store/storageEngineHooks';
+import { useAuth } from './store/hooks/useAuth';
 
 interface ProtectedRouteProps {
   children: ReactNode;
 }
 
 export function ProtectedRoute({ children }: ProtectedRouteProps) {
-  const { user, login } = useAuth();
-  const { storageEngine } = useStorageEngine();
-
-  useEffect(() => {
-    if (storageEngine?.getEngine() !== 'firebase') {
-      // If not using firebase as storage engine, authenticate the user with a fake user with admin privileges.
-      const currUser: User = {
-        name: 'localName',
-        admin: true,
-        email: 'localEmail@example.com',
-      };
-      login(currUser);
-    }
-  }, []);
+  const { user } = useAuth();
 
   if (user && user.admin) {
     return <div>{children}</div>;
--- a/src/ProtectedRoute.tsx
+++ b/src/ProtectedRoute.tsx
@@ -1,0 +1,17 @@
+import { useState, useEffect } from 'react';
+import { Navigate, Outlet } from 'react-router-dom';
+
+export interface User {
+  email: string;
+  admin: boolean;
+  name?: string;
+}
+
+interface ProtectedRouteProps {
+  user: User | undefined;
+  isAuth: boolean;
+}
+
+export function ProtectedRoute({ user, isAuth }: ProtectedRouteProps) {
+  return isAuth ? <Outlet /> : <Navigate to="/login" />;
+}
--- a/src/parser/types.ts
+++ b/src/parser/types.ts
@@ -18,6 +18,8 @@ export interface GlobalConfig {
   };
   /** A required property that is used to generate the list of available studies in the UI. This list is displayed on the landing page when running the app. */
   configsList: string[];
+  /** A required property listing out the emails of users who require admin access. */
+  adminUsers: string[];
 }
 
 /**
--- a/src/store/hooks/useAuth.tsx
+++ b/src/store/hooks/useAuth.tsx
@@ -1,0 +1,49 @@
+import {
+  createContext, useContext, useMemo, useState, ReactNode,
+} from 'react';
+import { useNavigate } from 'react-router-dom';
+
+export interface User {
+  name?: string|null;
+  email: string|null;
+  admin: boolean;
+}
+
+interface AuthContextValue {
+  user: User | null;
+  login: (user: User) => void;
+  logout: () => void;
+  }
+
+const AuthContext = createContext<AuthContextValue>({
+  user: null,
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
+  login: () => {},
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
+  logout: () => {},
+});
+
+export const useAuth = () => useContext(AuthContext);
+
+export function AuthProvider({ children } : { children: ReactNode }) {
+  const [user, setUser] = useState<User|null>(null);
+  const navigate = useNavigate();
+
+  const login = async (currUser: User) => {
+    setUser(currUser);
+    navigate('/');
+  };
+
+  const logout = () => {
+    setUser(null);
+    navigate('/login');
+  };
+
+  const value = useMemo(() => ({
+    user,
+    login,
+    logout,
+  }), [user]);
+
+  return <AuthContext.Provider value={value}>{children}</AuthContext.Provider>;
+}
--- a/src/store/hooks/useAuth.tsx
+++ b/src/store/hooks/useAuth.tsx
@@ -2,6 +2,7 @@ import {
   createContext, useContext, useMemo, useState, ReactNode,
 } from 'react';
 import { useNavigate } from 'react-router-dom';
+import { useLocalStorage } from './useLocalStorage';
 
 export interface User {
   name?: string|null;
@@ -26,7 +27,7 @@ const AuthContext = createContext<AuthContextValue>({
 export const useAuth = () => useContext(AuthContext);
 
 export function AuthProvider({ children } : { children: ReactNode }) {
-  const [user, setUser] = useState<User|null>(null);
+  const [user, setUser] = useLocalStorage('user', null);
   const navigate = useNavigate();
 
   const login = async (currUser: User) => {
--- a/index.html
+++ b/index.html
@@ -6,6 +6,7 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <title>reVISit</title>
 
+    
     <!-- Required for react browser router on GitHub pages -->
     <script type="text/javascript">
       // Single Page Apps for GitHub Pages
--- a/src/GlobalConfigParser.tsx
+++ b/src/GlobalConfigParser.tsx
@@ -8,7 +8,6 @@ import { AnalysisInterface } from './analysis/AnalysisInterface';
 import { PREFIX } from './utils/Prefix';
 import { ProtectedRoute } from './ProtectedRoute';
 import { Login } from './Login';
-import { StorageEngine } from './storage/engines/StorageEngine';
 import { AuthProvider } from './store/hooks/useAuth';
 
 async function fetchGlobalConfigArray() {
@@ -33,7 +32,7 @@ async function fetchStudyConfigs(globalConfig: GlobalConfig) {
   return studyConfigs;
 }
 
-export function GlobalConfigParser({ storageEngine } : { storageEngine:StorageEngine|undefined }) {
+export function GlobalConfigParser() {
   const [globalConfig, setGlobalConfig] = useState<Nullable<GlobalConfig>>(null);
   const [studyConfigs, setStudyConfigs] = useState<Record<string, StudyConfig>>({});
 
@@ -73,16 +72,20 @@ export function GlobalConfigParser({ storageEngine } : { storageEngine:StorageEn
             path="/:studyId/*"
             element={<Shell globalConfig={globalConfig} />}
           />
-
           <Route
             path="/analysis/:page"
-            element={<AnalysisInterface globalConfig={globalConfig} />}
+            element={(
+              <ProtectedRoute>
+                <AnalysisInterface
+                  globalConfig={globalConfig}
+                />
+              </ProtectedRoute>
+            )}
           />
           <Route
             path="/login"
             element={(
               <Login
-                storageEngine={storageEngine}
                 admins={globalConfig.adminUsers}
               />
             )}
--- a/src/GlobalConfigParser.tsx
+++ b/src/GlobalConfigParser.tsx
@@ -6,9 +6,10 @@ import { parseGlobalConfig, parseStudyConfig } from './parser/parser';
 import { GlobalConfig, Nullable, StudyConfig } from './parser/types';
 import { AnalysisInterface } from './analysis/AnalysisInterface';
 import { PREFIX } from './utils/Prefix';
-import { ProtectedRoute, User } from './ProtectedRoute';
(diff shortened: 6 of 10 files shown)


src/parser/types.ts
```diff
@@ -18,6 +18,8 @@ export interface GlobalConfig {
   };
   /** A required property that is used to generate the list of available studies in the UI. This list is displayed on the landing page when running the app. */
   configsList: string[];
+  /** A required property listing out the emails of users who require admin access. */
+  adminUsers: string[];
```
JackWilb, line 22: Should this be required? We could disable if not provided

src/storage/engines/StorageEngine.ts
```diff
@@ -1,3 +1,4 @@
+import { DocumentData, DocumentReference } from 'firebase/firestore';
```
JackWilb, line 1: This should be in the firebase storage engine, right?

src/storage/engines/FirebaseStorageEngine.ts
```diff
@@ -376,6 +376,30 @@ export class FirebaseStorageEngine extends StorageEngine {
     return allAnswersPresent;
   }
 
+  async getUserManagementData(key:string) {
+    // Get the user-management collection in Firestore
+    const userManagementCollection = collection(this.firestore, 'user-management');
+    // Grabs all user-management data and returns data based on key
+    const userManagementDocs = await getDocs(userManagementCollection)
+      .then((querySnapshot) => {
```
JackWilb, line 384: You're using async/await so you shouldn't use then. You should await the thing that you need to wait on and use the result in the logic after

JackWilb (commented):
Some initial comments. I'll have some more for you after an iteration

src/store/hooks/useAuth.tsx
```diff
@@ -0,0 +1,223 @@
...
+
+type verifyAdminStatusInput = User | null | LocalStorageUser;
+
+// Defines default AuthContextValue
+interface AuthContextValue {
+  user: UserWrapped;
+  adminVerification:boolean;
+  login: (user: UserWrapped) => void;
+  logout: () => void;
+  verifyAdminStatus: (firebaseUser:verifyAdminStatusInput) => Promise<boolean>;
```
JackWilb, line 32:
```suggestion
  verifyAdminStatus: (firebaseUser: UserOptions) => Promise<boolean>;
```

src/store/hooks/useAuth.tsx
```diff
@@ -0,0 +1,223 @@
...
+}
+
+type verifyAdminStatusInput = User | null | LocalStorageUser;
+
+// Defines default AuthContextValue
+interface AuthContextValue {
+  user: UserWrapped;
+  adminVerification:boolean;
+  login: (user: UserWrapped) => void;
+  logout: () => void;
```
JackWilb, line 31: Since these modify state they're async

```suggestion
  login: (user: UserWrapped) => Promise<void>;
  logout: () => Promise<void>;
```

src/store/hooks/useAuth.tsx
```diff
@@ -0,0 +1,223 @@
...
+  );
+
+  const [adminVerification, setAdminVerification] = useState<boolean>(false);
+
+  const { storageEngine } = useStorageEngine();
+
+  // Logs in the user (i.e. sets the user info and navigates to the root)
+  const login = (inputUser: UserWrapped) => {
+    setUser(inputUser);
+  };
```
JackWilb, line 71: You'll need to play with this, but this should allow you to await the login. You'll need something similar for logout to await that too

```suggestion
  const login = async (inputUser: UserWrapped) => {
    return new Promise((resolve, reject) => {
      setUser(inputUser, () => resolve());
    }
  };
```

src/store/hooks/useAuth.tsx
```diff
@@ -0,0 +1,223 @@
...
+      // Assert that the inputted user is the User type
+      const firebaseUser = inputUser as User;
+      if (storageEngine instanceof FirebaseStorageEngine) {
+        const isAuthEnabled = await storageEngine?.getUserManagementData('authentication').then((data) => data?.isEnabled);
+        if (isAuthEnabled !== false) {
+          // Validate the user if Auth enabled
+          return await storageEngine.validateUserAdminStatus(firebaseUser, globalConfig.adminUsers);
+        }
+        // If authDisabled, allow access
+        return Promise.resolve(true);
```
JackWilb, line 119:
```suggestion
        return true;
```

src/store/hooks/useAuth.tsx
```diff
@@ -0,0 +1,223 @@
...
+      }
+      // If the user is Firebase but the storageEngine is not, something has gone wrong. Deny access
+      return Promise.resolve(false);
```
JackWilb, line 122:
```suggestion
      return false;
```

src/store/hooks/useAuth.tsx
```diff
@@ -0,0 +1,223 @@
...
+    } if (isLocalStorageUser(inputUser)) {
+      if (storageEngine) {
+        // If storageEngine defined, allow access
+        return Promise.resolve(true);
+      }
+      // If localStorageUser but no storageEngine defined, deny access.
+      return Promise.resolve(false);
+    }
+    // Some other forms of the user input should be denied access.
+    return Promise.resolve(false);
```
JackWilb, line 132:
```suggestion
        return true;
      }
      // If localStorageUser but no storageEngine defined, deny access.
      return false;
    }
    // Some other forms of the user input should be denied access.
    return false;
```

src/store/hooks/useAuth.tsx
```diff
@@ -0,0 +1,223 @@
...
+      });
+      if (firebaseUser) {
+        // Reach out to firebase to validate user
+        const isAdmin = await verifyAdminStatus(firebaseUser);
+        setAdminVerification(true);
+        login({
+          user: firebaseUser,
+          determiningStatus: false,
+          isAdmin,
+        });
```
JackWilb, line 167: Setting the user async here is making me a little nervous

src/store/hooks/useAuth.tsx
```diff
@@ -0,0 +1,223 @@
...
+      } else {
+        logout();
+      }
+    };
+
+    // Determine authentication listener based on storageEngine and authEnabled variable
+    const determineAuthentication = async () => {
+      if (storageEngine instanceof FirebaseStorageEngine) {
+        const isAuthEnabled = await storageEngine?.getUserManagementData('authentication').then((data) => data?.isEnabled);
```
JackWilb, line 176:
```suggestion
        const authInfo = await storageEngine?.getUserManagementData('authentication');
        const isAuthEnable = authInfo?.isEnabled;
```

src/store/hooks/useAuth.tsx
```diff
@@ -0,0 +1,223 @@
...
+          determiningStatus: false,
+          isAdmin,
+        });
+      } else {
+        logout();
+      }
+    };
+
+    // Determine authentication listener based on storageEngine and authEnabled variable
+    const determineAuthentication = async () => {
```
JackWilb, line 174: Do we need this?

src/storage/engines/FirebaseStorageEngine.ts
```diff
@@ -18,6 +18,11 @@ import { hash } from './utils';
 import { StudyConfig } from '../../parser/types';
 import { getSequenceFlatMap } from '../../utils/getSequenceFlatMap';
 
+export interface StoredUser {
+  email:string,
+  uid:string|null,
```
JackWilb, line 23:
```suggestion
  email: string,
  uid: string | null,
```

src/storage/engines/FirebaseStorageEngine.ts
```diff
@@ -376,6 +389,96 @@ export class FirebaseStorageEngine extends StorageEngine {
     return allAnswersPresent;
   }
 
+  async getUserManagementData(key:string) {
+    // Get the user-management collection in Firestore
+    const userManagementCollection = collection(this.firestore, 'user-management');
+    // Grabs all user-management data and returns data based on key
+    const userManagementDocs = await getDocs(userManagementCollection)
+      .then((querySnapshot) => {
```
JackWilb, line 397: Please don't use `await getDocs().then()`

src/storage/engines/FirebaseStorageEngine.ts
```diff
@@ -376,6 +389,96 @@ export class FirebaseStorageEngine extends StorageEngine {
...
+        const docsObject:Record<string, DocumentData> = {};
```
JackWilb, line 398:
```suggestion
        const docsObject: Record<string, DocumentData> = {};
```

src/storage/engines/FirebaseStorageEngine.ts
```diff
@@ -376,6 +389,96 @@ export class FirebaseStorageEngine extends StorageEngine {
...
+        querySnapshot.docs.forEach((queryDoc) => {
+          docsObject[queryDoc.id] = queryDoc.data();
+        });
```
JackWilb, line 401: This should be a map since you're assigning.

E.g.: `const docsObject = Object.fromEntries(querySnapshot.docs.map((queryDoc) => [queryDoc.id, queryDoc.data()]));`

public/global.json
```diff
@@ -51,5 +51,8 @@
         "demo-image": {
             "path": "demo-image/config.json"
         }
-    }
+    },
+    "adminUsers":[
+        "briancbollen@gmail.com"
+    ]
```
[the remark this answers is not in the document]
bbollen23: Yeah, mistakenly added. Will fix that. I think by default though we'll have to have it blank because of the logic. Otherwise, it will try to add that email I believe and then you'll be locked out.

src/store/hooks/useAuth.tsx
bbollen23: Ahhh, this makes a lot of sense. Thank you!!

src/components/ConfigSwitcher.tsx
```diff
@@ -16,9 +18,19 @@ type Props = {
...
   const { configsList } = globalConfig;
   const navigate = useNavigate();
+  const { logout } = useAuth();
+  const { storageEngine } = useStorageEngine();
 
   return (
     <Container size="xs" px="xs" style={{ marginTop: 100, marginBottom: 100 }}>
+      {storageEngine?.getEngine() === 'firebase' ? (
+        <UnstyledButton onClick={() => {
+          logout(); navigate('/login');
```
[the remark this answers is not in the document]
bbollen23: Yes, I also don't think that it's even necessary. If we logout on a protected route, it'll revalidate the admin status and kick them out.

src/store/hooks/useAuth.tsx
```diff
@@ -0,0 +1,223 @@
...
+
+  const verifyAdminStatus = async (inputUser: verifyAdminStatusInput) => {
+    if (inputUser === null) {
+      return Promise.resolve(true);
+    } if (isFirebaseUser(inputUser)) {
+      // Assert that the inputted user is the User type
+      const firebaseUser = inputUser as User;
+      if (storageEngine instanceof FirebaseStorageEngine) {
+        const isAuthEnabled = await storageEngine?.getUserManagementData('authentication').then((data) => data?.isEnabled);
+        if (isAuthEnabled !== false) {
```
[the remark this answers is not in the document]
bbollen23: My general idea is that auth is enabled even if the user hasn't added the authInfo to the userManagement collection. So it's only disabled when it's explicitly false.

src/store/hooks/useAuth.tsx
bbollen23: This is meant to be the callback for the onAuthStatusChanged function. So I don't think the asynchronous behavior is an issue -- it's a listener which, whenever the auth status changes, the user gets reverified. I am going to remove the initial reset of the user though. We don't want the user to get reset if they are indeed an admin and their refresh token just expired. Instead, this'll be changed to just change the determining status of the user.

src/store/hooks/useAuth.tsx
```diff
@@ -0,0 +1,212 @@
...
+  // Non-auth User
+  const nonAuthUser : UserWrapped = {
+    user: {
+      name: 'fakeName',
+      email: 'fakeEmail@fake.com',
+    },
+    determiningStatus: false,
+    isAdmin: true,
+    adminVerification: true,
+  };
```
JackWilb, line 87: A way to use the user objects that you create

```suggestion
  // Default non-user when loading
  const loadingNullUser : UserWrapped = {
    user: null,
    determiningStatus: true,
    isAdmin: false,
    adminVerification: false,
  };

  // Default non-user when not loading
  const nonLoadingNullUser : UserWrapped = {
    user: null,
    determiningStatus: false,
    isAdmin: false,
    adminVerification: false,
  };

  // Non-auth User
  const nonAuthUser : UserWrapped = {
    user: {
      name: 'fakeName',
      email: 'fakeEmail@fake.com',
    },
    determiningStatus: false,
    isAdmin: true,
    adminVerification: true,
  };
  
  const [user, setUser] = useState(loadingNullUser);
```

src/store/hooks/useAuth.tsx
```diff
@@ -0,0 +1,212 @@
...
+  user: UserOptions,
+  determiningStatus: boolean,
+  isAdmin: boolean,
+  adminVerification:boolean
+}
+
+// Defines default AuthContextValue
+interface AuthContextValue {
+  user: UserWrapped;
+  logout: () => void;
```
JackWilb, line 30: `logout` is `async` below. Do we need to modify the type here?

src/store/hooks/useAuth.tsx
```diff
@@ -0,0 +1,212 @@
...
+  // Logs the user out by removing the user and navigating to '/login'
+  const logout = async () => {
+    const auth = getAuth();
+    try {
+      await signOut(auth);
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    } catch (error: any) {
+      console.error(`There was an issue signing-out the user: ${error.message}`);
+    }
+    setUser(nonLoadingNullUser);
```
JackWilb, line 100:
```suggestion
    try {
      await signOut(auth);
    // eslint-disable-next-line @typescript-eslint/no-explicit-any
    } catch (error: any) {
      console.error(`There was an issue signing-out the user: ${error.message}`);
    } finally {
      setUser(nonLoadingNullUser);
    }
```

src/store/hooks/useAuth.tsx
```diff
@@ -0,0 +1,212 @@
...
+
+  const verifyAdminStatus = async (inputUser: UserOptions) => {
+    if (inputUser === null) {
+      return true;
+    } if (isFirebaseUser(inputUser)) {
+      // Assert that the inputted user is the User type
+      const firebaseUser = inputUser;
+      if (storageEngine instanceof FirebaseStorageEngine) {
+        const authInfo = await storageEngine?.getUserManagementData('authentication');
+        const isAuthEnabled = authInfo !== null ? authInfo.isEnabled : true;
```
JackWilb, line 119: I think you also need to check undefined here, so we can just remove the explicit null check

```suggestion
        const isAuthEnabled = authInfo ? authInfo.isEnabled : true;
```

JackWilb (approved):
Approved with minor changes

src/ProtectedRoute.tsx
```diff
@@ -0,0 +1,43 @@
...
+  }, [user.isAdmin]);
+
+  if (user.determiningStatus || !storageEngine) {
+    return <LoadingOverlay overlayOpacity={0} visible={user.determiningStatus || !storageEngine?.getEngine()} />;
+  }
+
+  if (!user.isAdmin && !user.determiningStatus) {
+    return <Navigate to="/login" />;
+  }
+  return <div>{children}</div>;
```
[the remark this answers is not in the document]
bbollen23: I don't think it's going to make much of a difference, but stylistically it should definitely be the fragment as you suggested (then we don't have to worry :) )

src/storage/engines/FirebaseStorageEngine.ts
```diff
@@ -376,6 +389,95 @@ export class FirebaseStorageEngine extends StorageEngine {
...
+        const isAdmin = user.email && (adminUsersObject[user.email] === user.uid || adminUsersObject[user.email] === null);
+        if (isAdmin) {
+          await this.editUserManagementAdmins(globalConfigAdminUsers, user);
+          return true;
+        }
+        return false;
+      // Case 2: Database does not yet exist. First opening
+      }
+      // Need to get Global config users
+      const isAdmin = (user.email?.includes && globalConfigAdminUsers.includes(user.email)) ?? false;
```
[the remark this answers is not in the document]
bbollen23: Just verified this -- yes, not necessary. Believe this was mistakenly left over when writing this differently (i.e. doing globalConfigAdminusers.includes && globalConfigAdminUsers.includes(user.email) or something to that effect)

src/store/hooks/useAuth.tsx
bbollen23: The compiler isn't having a problem with it, but not sure why tbh. I changed to the following:

In  interface AuthContextValue: 
From `logout: () => void` to `logout: () => Promise<void>`

In Auth Context initialization:
From `logout: () => {};` to `logout: async () => {};`

src/components/ConfigSwitcher.tsx
```diff
@@ -16,9 +18,21 @@ type Props = {
...
+      {storageEngine?.getEngine() === 'firebase' ? (
+        <UnstyledButton
+          style={{ position: 'absolute', top: '30px', right: '30px' }}
+          onClick={() => {
+            logout();
+          }}
+        >
+          Logout
+        </UnstyledButton>
+      ) : null}
```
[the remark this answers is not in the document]
bbollen23: Issue here: https://github.com/revisit-studies/study/issues/299

PR #295, opened 2024-04-13
286 website component documentation

### LINKED ISSUE #286: Website Component Documentation
We need some additional information on the Website Component type. I think we should do the following:
1) Show a simple example of using a website with no communication
2) Show how we use the revisit-communicate.js to communicate between reVISit. For now, I think this would be a good place to do very explicit documentation of the functions within this file until we decide a better location.
3) Any other general information about this component that would be useful.

Check out how we have been doing the documentation for the StudyConfig, UIConfig, and StudyMetada types for syntax and such.

hilsonshrestha:
Closes #286 

Add documentation for website component and passing data between reVISit and the website/html file.

x

x


src/parser/types.ts
```diff
@@ -341,6 +341,38 @@ export interface ImageComponent extends BaseIndividualComponent {
...
+  <script src="../../revisitUtilities/revisit-communicate.js"></script>
+  <script>
+    Revisit.onDataReceive((data) => {
+      const barData = data['barData'];
+      ...
+    });
+
+    Revisit.postAnswers({ answer: [userAnswer] });
+  </script>
+```
```
bbollen23, line 375: This is great! Can you go a little more in depth on how "postAnswers" works? I think it's clear how we can pull data in using the onDataReceive, but what is the object that we pass to postAnswers and why? Does it always have to be  "({answer: [userAnswer]})"? I think it would be good use the postAnswers as an example within an "onClick" action to describe more explicitly how we can use it.

public/revisitUtilities/revisit-communicate.js
```diff
@@ -43,13 +43,6 @@
         documentWidth: document.documentElement.scrollWidth,
       });
     },
-    // Inform Revisit that the stimuli is ready in the iframe.
-    postReady: () => {
-      sendMessage('READY', {
-        documentHeight: document.documentElement.scrollHeight,
-        documentWidth: document.documentElement.scrollWidth,
-      });
-    },
```
bbollen23, line 52: Thank you for cleaning this up

src/parser/StudyConfigSchema.json
bbollen23, line 1: Have you verified that this renders properly? I'm personally not sure how this would work. Let me know

bbollen23 (changes requested):

src/parser/types.ts
hilsonshrestha: I need to check on this with Jack. I believe originally I had just a single answer, but was changed later for provenance.

src/parser/StudyConfigSchema.json
hilsonshrestha: Nope, it does not render as intended. But, we do have to run `yarn run generate-schema:StudyConfig` whenever we have changes to the types.ts. This generates the StudyConfigSchema.json

I think the schema only supports plain text and does not support code snippet and stuffs which we have added in types.ts.

bbollen23: Oh, so the changes in the StudyConfigSchema.json were generated by the `yarn run generate-schema:StudyConfig` and not by you personally?

hilsonshrestha: yes that's right. It was generated by running that script.

src/parser/types.ts
bbollen23: Okay, just keep this updated with what you find.

hilsonshrestha: I have updated the data structure and the example to just send the answers. Also, I have a separated the method to send provenance data. I have create a separate issue for it https://github.com/revisit-studies/study/issues/321

src/parser/types.ts
```diff
@@ -380,6 +380,47 @@ export interface ImageComponent extends BaseIndividualComponent {
 
 /**
  * The WebsiteComponent interface is used to define the properties of a website component. A WebsiteComponent is used to render an iframe with a website inside of it. This can be used to display an external website or an html file that is located in the public folder.
+
+``` JSON
+  {
+    "type": "website",
+    "path": "path/to/website.html",
```
bbollen23, line 387:
```suggestion
    "path": "path/to/study/assets/website.html",
```

src/parser/types.ts
```diff
@@ -380,6 +380,47 @@ export interface ImageComponent extends BaseIndividualComponent {
...
+
+``` HTML
+  <script src="../../revisitUtilities/revisit-communicate.js"></script>
+  <script>
+    Revisit.onDataReceive((data) => {
+      const barData = data['barData'];
+      ...
+    });
+
+    Revisit.postAnswers({ barChart: userAnswer });
```
bbollen23, line 421: Call out that 'barChart' needs to match ID in 'response' object

bbollen23 (changes requested):

PR #297, opened 2024-04-22
Adding docs for the react component

### LINKED ISSUE #283: React Component Documentation
We need some additional information on the React Component type. I suggest adding details about 1) location of components, 2) the paths in the StudyConfig being relative to `src/public/` rather than `public/`, 3) a simple example showcasing how it works, 4) maybe a more in-depth example. Possibly point to specific demos using these as well. Check out how we have been doing the documentation for the StudyConfig, UIConfig, and StudyMetada types for syntax and such.

ZachCutler04:
Closes #283 

Adds simple docs to the react component explaining the basics, and links to relevant examples. 
I'm not actually sure if the typedoc will pick up on the code snippets I have? Not sure if thats the right syntax


src/parser/types.ts
```diff
@@ -319,6 +319,44 @@ export interface MarkdownComponent extends BaseIndividualComponent {
...
+ *
+ * React components created this way have a generic prop type <StimulusParams<T>>, which has the following types.
+ * {
+ *  parameters: T;
+ *  setAnswer: ({ status, provenanceGraph, answers }: { status: boolean, provenanceGraph?: TrrackedProvenance, answers: Record<string, any> }) => void
+ * }
+ *
+ * parameters is the same object passed in from the ReactComponent type.
+ * setAnswer is a callback function allowing the creator of the ReactComponent to programmatically set the answer, as well as the provenance graph. This can be useful if you dont use the default answer interface, and instead have something more unique.
+ * See the click-accuracy-test link below for an example of this.
```
bbollen23, line 334: This feels a bit confusing to me. On first read, I can't personally understand what this section is describing. I think providing an example of what this is doing would make more sense/provide a better flow. Our goal here is to provide details but making sure these details are rooted in concrete examples. I understand that you're pointing to the 'click-accuracy-test', but I think it would be better to possibly bring in the snippet of code from there. 

I think this example could also be moved to the bottom after you've already given the basics. And maybe lead in with something like "Like our other components, React components come with a default answer interface. However, we do have support for capturing more complex stimuli in React components..."

bbollen23 (dismissed):
In addition to these comments, I think some things may be better placed at above the actual parameter types below all of this (like the path stuff can just go about the path parameter).

JackWilb (approved):

PR #322, opened 2024-04-27
Add a completed variable to the participant data to find when people finish studies (req. for skip logic changes)

JackWilb:
No

This fixes a bug with the analysis dashboard and with the data. Now that we're allowing skipping, we need to record when a participant has finished the study. I'm adding a `completed` variable to the participant data that is updated when we `verifyAnswers`.

N/A

No


ZachCutler04 (commented):
looks good

ZachCutler04 (approved):

PR #355, opened 2024-05-06
Fixed UIDs not being added

bbollen23:
When an admin is added, the uid is defaulted to be `null`. This was intended to be set to the correct value once that user signs in for the first time. This functionality was removed by mistake when refactoring code and has been now added into the validateUser function for the Firebase storage engine.


src/storage/engines/FirebaseStorageEngine.ts
```diff
@@ -424,6 +424,17 @@ export class FirebaseStorageEngine extends StorageEngine {
           // Verifies that, if the user has signed in and thus their UID is added to the Firestore, that the current UID matches the Firestore entries UID. Prevents impersonation (otherwise, users would be able to alter email to impersonate).
           const isAdmin = user.user.email && (adminUsersObject[user.user.email] === user.user.uid || adminUsersObject[user.user.email] === null);
           if (isAdmin) {
+            // Add UID to user in collection if not existent.
+            if (user.user.email && adminUsersObject[user.user.email] === null) {
+              for (let i = 0; i < adminUsers.adminUsersList.length; i += 1) {
+                if (adminUsers.adminUsersList[i].email === user.user.email) {
+                  adminUsers.adminUsersList[i].uid = user.user.uid;
+                }
+              }
```
JackWilb, line 433: This is a bit more in line with javascript ES6 style

```suggestion
              const adminUser = adminUsers.adminUsersList.find((u) => u.user.email === user.user.email);
              if (adminUser) {
                adminUser.user.uid = user.user.uid
              }
```

JackWilb (approved):

Issue #266, opened 2024-03-11
Moving config to firebase storage


dyr429:
**Is your feature request related to a problem? Please describe.**
The config file is currently stored in the cloud fire store, when the config file exceeds the max allowed 1,048,576 bytes, it throws error.

**Describe the solution you'd like**
Moving config file to storage, like sequences.

**Additional context**
This error happens when we are developing an experiment that has 1400 questions.

JackWilb:
Prefix storage folders with dev- or prod-

JackWilb:
Closed via #387

Issue #253, opened 2024-02-20
Participant data can become too large for real time documents


JackWilb:
We should move the participant answers to storage so they never grow too large for the realtime database.

JackWilb:
Closed via #387

PR #387, opened 2024-06-03
Clear database feature

### LINKED ISSUE #334: Clear out database in admin interface
Via GUI. 
remove participant data, sequence array, etc.
Make sure this is "safe" from a UI perspective, similar to "delete repo" in github

### LINKED ISSUE #384: Google Login for first Auth: "The requested action is invalid."
**Describe the bug**
Background:
I encountered this bug on the 21st May 2024, I haven't merged my fork since then.
When following the tutorial, after completing the 'Authentication' step and attempting to add the first user as the Admin via Study browser > Settings, the login popup returns the error "The requested action is invalid.". When inspecting the popup with the browsers devtools, the console contains the error "Request is missing required data". I tried different browsers, computers and disabling adblockers.

**Potential Fix:**
Forewarning in that I know next to nothing about Firebase, so this may not be the best solution.
I fixed this issue by adding a resolver to the function `signInWithPopup()` in `./src/Login.tsx`. Specifically, I:

```
// Added an import of browserPopupRedirectResolver from @firebase/auth
import {
  getAuth, signInWithPopup, GoogleAuthProvider,
  browserPopupRedirectResolver,
} from '@firebase/auth';

//In the signInWithGoogle() function block
// Added browserPopupRedirectResolver to signInWithPopup() as the third param.
await signInWithPopup(auth, provider, browserPopupRedirectResolver);
```

An example can be seen here: https://github.com/broldfield/revisit/blob/main/src/Login.tsx

After adding this, the Auth begun working as expected. I haven't tried reproducing the bug after adding in the fix as I don't want to mess with `user-management` in firebase.

### LINKED ISSUE #386: All Storage directories should use collection prefix
**Problem**
Data stored in Firebase storage (not Firestore) does not prefix the datasets with a "dev-" or "prod-" string. When we store data in Firestore, this is handled correctly.

**Solution**
When accessing and pushing data to storage, we need to prefix with ${this.collectionPrefix}

bbollen23:
Closes https://github.com/revisit-studies/study/issues/266
Closes https://github.com/revisit-studies/study/issues/253
Closes #334 
Closes #386 
Closes #384 

This allows for better data management when using Firebase. What created an additional tab in the individual analysis views which is meant for data management. The functionality is as follows:

1. A user can "archive" a dataset. When you archive the data, it is copied to a separate directory in storage labeled exactly the same as the study with an additional "-archive-YYYYMMDDhhmmss" suffix. 
2. A user can "archive and delete" a dataset. This will not only archive the data, but will clear out the current study data as well. We can think of this as a "reset" and we can think of just archiving as a "snapshot".
3. A user can "restore" any archive. When you restore an archive, it will first archive the current study data if it exists. Then, it will copy all data from the archive to be restored into the current study data. The archive that is restored will then be removed. 
4. A user can delete any archive. The list of archives is on the same page.

Screenshot 2024-06-03 at 1 05 03 PM (2)

TODOs:
- [x] Change to 'confirm modals'

### DIFF
--- a/src/storage/engines/FirebaseStorageEngine.ts
+++ b/src/storage/engines/FirebaseStorageEngine.ts
@@ -535,6 +535,8 @@ export class FirebaseStorageEngine extends StorageEngine {
   }
 
   async removeSnapshot(targetName:string) {
+    await this._deleteDirectory(`${targetName}/configs`);
+    await this._deleteDirectory(`${targetName}/participants`);
     await this._deleteDirectory(targetName);
     await this._deleteCollection(targetName);
     await this._removeNameFromMetadata(targetName);
--- a/src/storage/engines/FirebaseStorageEngine.ts
+++ b/src/storage/engines/FirebaseStorageEngine.ts
@@ -7,20 +7,34 @@ import {
   ref,
   uploadBytes,
   listAll,
+  FirebaseStorage,
+  StorageReference,
 } from 'firebase/storage';
 import {
-  CollectionReference, DocumentData, Firestore, collection, doc, enableNetwork, getDoc, getDocs, initializeFirestore, orderBy, query, serverTimestamp, setDoc, where, deleteDoc, updateDoc, writeBatch,
+  CollectionReference, DocumentData, Firestore, collection, doc, enableNetwork, getDoc, getDocs, initializeFirestore, orderBy, query, serverTimestamp, setDoc, where, deleteDoc, updateDoc,
 } from 'firebase/firestore';
 import { ReCaptchaV3Provider, initializeAppCheck } from '@firebase/app-check';
 import { getAuth, signInAnonymously } from '@firebase/auth';
 import localforage from 'localforage';
+import { v4 as uuidv4 } from 'uuid';
 import { StorageEngine, UserWrapped, StoredUser } from './StorageEngine';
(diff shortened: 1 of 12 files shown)


src/storage/engines/FirebaseStorageEngine.ts
```diff
@@ -555,32 +499,257 @@ export class FirebaseStorageEngine extends StorageEngine {
...
+      const response = await fetch(sourceFileURL);
+      const blob = await response.blob();
+
+      await uploadBytes(targetFileRef, blob);
+    } catch (error) {
+      console.error('Error copying file:', error);
+    }
+  }
+
+  async _copyCollection(sourceCollectionName: string, targetCollectionName: string) {
```
JackWilb, line 651: These `_{copy,delete}collection` methods are going to sting us when we add anything else to the firestore... I think we need to make these more general, but I don't want to slow this PR down. Let's make an issue for these for when we have some extra cycles

JackWilb (approved):

src/storage/engines/FirebaseStorageEngine.ts
bbollen23: Yeah, I agree. It was simpler to do this rather than make a recursive one with all these edge cases since we really only copy this. I'll make an issue for it

PR #390, opened 2024-06-06
Porting to mantine 7

ZachCutler04:
#368 

Updating everything to mantine 7. Mostly just prop changes, only real code changes are ensuring there is a correct AppShell/AppShell.{whatever} around all of the components. This should have been happening already, but is now very required in mantine 7.

Theoretically everything looks about the same, the most noticeable differences are in the form elements, which we can consider changing to look more like they used to (I think they looked better before)


src/components/ReactMarkdownWrapper.tsx
```diff
@@ -9,7 +9,7 @@ import rehypeRaw from 'rehype-raw';
 
 export default function ReactMarkdownWrapper({ text, required }: { text: string; required?: boolean }) {
   const components = {
-    img: ({ node, ...props }: { node: unknown; }) => <Image {...props} />,
+    img: ({ node, ...props }: { node: unknown; height?: string | number, width?: string | number }) => <Image {...props} h={props.height} w={props.width} />,
```
JackWilb, line 12: Isn't this passed in as style?

JackWilb (approved):

PR #447, opened 2024-08-07
Snapshot renaming and error catching

### LINKED ISSUE #446: Snapshots should be disabled with no data
A user should not be able to try and take a snapshot/archive data if no data exists yet. Additionally, we should show errors when failing to archive or create snapshots.

bbollen23:
Closes #446 , #445 

This PR handles the following:
- Added general notification system. Currently used only by FirebaseEngine. Can be accessed by importing "showNotification" and "RevisitNotification" type from the "notifications.ts" file in "utils/".
- When trying to create a snapshot or archive, if there is no data in the current study, an error will be shown as a notification.
- If desired, we can now use the "FirebaseActionResponse" interface in the Firebase storage engine class to assert that the response from any Firebase action will have a status of "SUCCESS" or "FAILED", with a required "error" when "FAILED". "SUCCESS" has an optional set of "notifications" that can then be rendered in our component if desired. This is useful for trying to create snapshots and may be useful for other Firebase actions (designed to be similar to an API response).
- Snapshots can now be renamed. Edit icon has been added to left of "restore". Opens a modal for renaming
- All snapshots maintain the original snapshot order (so it is always in date created order).
- Notifications are stacked. For example, when we restore a snapshot, two actions are performed: first, a new snapshot of the live data is created then the restored data will be copied into the live data. However, if no live data exists, then a warning notification will be shown displaying this. But, the snapshot to be restored will still be copied into the live data (so not a full error, just a partial one). Similarly with archiving: if an error occurs when deleting live data, the notification will be shown displaying this error, but it will not prevent a snapshot from being created.
- Changed snapshots list to be a table view. Date Created column has been added.

Screenshot 2024-08-06 at 7 08 07 PM
Screenshot 2024-08-06 at 7 08 16 PM
**NOTE:** The messaging for failing to create a snapshot has been changed in the actual code from the image you see below. 
Screenshot 2024-08-06 at 7 08 16 PM
Screenshot 2024-08-06 at 7 08 16 PM
Screenshot 2024-08-06 at 7 08 16 PM
Screenshot 2024-08-08 at 9 08 46 AM
Screenshot 2024-08-08 at 9 08 46 AM

TODOs:
- [ ] Want to add some documentation which points out that upon deletion of live data (or archiving), the config will be re-fetched. So the user will always see one config in the Firebase UI
- [x] Need to fix loading with too many snapshots.
- [X] Should also include some sort of "data successfully restored" or something when you restore snapshots. Maybe just a non modal alert?
- [X] Double check what happens when you restore and there is no live data


src/analysis/individualStudy/management/DataManagementAccordionItem.tsx
```diff
@@ -37,20 +44,32 @@ export function DataManagementAccordionItem({ studyId, refresh }: { studyId: str
...
+    const createdSnapshot: FirebaseActionResponse = await storageEngine.createSnapshot(studyId, true);
+    if (createdSnapshot.status === 'SUCCESS') {
+      refreshSnapshots();
+      setLoading(false);
+      await refresh();
+    } else {
+      setLoading(false);
+      setError(createdSnapshot.error);
+      setModalErrorOpened(true);
+    }
```
JackWilb, line 72: Is it worth refactoring this repeated block into another function in this file?

src/analysis/individualStudy/management/DataManagementAccordionItem.tsx
```diff
@@ -266,6 +304,34 @@ export function DataManagementAccordionItem({ studyId, refresh }: { studyId: str
...
+        opened={modalRenameSnapshotOpened}
+        onClose={() => { setModalRenameSnapshotOpened(false); setRenameValue(''); }}
+        title={<Text>Rename Snapshot</Text>}
+      >
+        {/* <Text mb="sm">
+          This will permanently remove this snapshot. This action is
+          {' '}
+          <Text span fw={700}>irreversible</Text>
+          .
+        </Text> */}
```
JackWilb, line 317: To remove?

src/storage/engines/FirebaseStorageEngine.ts
```diff
@@ -614,12 +850,29 @@ export class FirebaseStorageEngine extends StorageEngine {
...
+  async renameSnapshot(directoryName: string, newName: string) {
+    try {
+      const metadataDoc = doc(this.firestore, 'metadata', 'collections');
+      await setDoc(
+        metadataDoc,
+        { [directoryName]: { enabled: true, name: newName } },
+        { merge: true },
+      );
+    } catch (error) {
+      console.error('Error renaming collection in metadata', error);
```
JackWilb, line 872: Should this be presented to the frontend with the new error handling you have?

JackWilb (approved):

src/storage/engines/FirebaseStorageEngine.ts
bbollen23: Might as well. If you like how I handled the previous one, I can do the same for this.

src/analysis/individualStudy/management/DataManagementAccordionItem.tsx
bbollen23: Thought about this too. Stuff is getting a little burdensome. I'll see how it feels when I try to refactor.

PR #452, opened 2024-08-09
Adding download snapshots

### LINKED ISSUE #451: Add downloading buttons to snapshots

bbollen23:
Closes #451 

Added download buttons to each snapshot. When clicked, the data is labeled with the snapshot name (the alternate name, so if it was renamed, it will use the renamed value).

To do this, the download buttons component needed to be adjusted to take in a list of participants data (like originally done) OR a function which fetches that data. It also takes in a couple other optional attributes for styling and file naming. 

Screenshot 2024-08-09 at 8 37 56 AM

### DIFF
--- a/src/analysis/individualStudy/management/DataManagementAccordionItem.tsx
+++ b/src/analysis/individualStudy/management/DataManagementAccordionItem.tsx
@@ -10,7 +10,6 @@ import { DownloadButtons } from '../../../components/downloader/DownloadButtons'
 import {
   FirebaseStorageEngine, FirebaseActionResponse, SnapshotNameItem,
 } from '../../../storage/engines/FirebaseStorageEngine';
-import { ParticipantData } from '../../../storage/types';
 
 export function DataManagementAccordionItem({ studyId, refresh }: { studyId: string, refresh: () => Promise<void> }) {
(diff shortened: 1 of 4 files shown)


src/analysis/individualStudy/management/DataManagementAccordionItem.tsx
```diff
@@ -131,6 +133,12 @@ export function DataManagementAccordionItem({ studyId, refresh }: { studyId: str
     }
     return null;
   };
+
+  const fetchParticipants = async (snapshotName: string): Promise<ParticipantData[]> => {
+    const strippedFilename = snapshotName.slice(snapshotName.indexOf('-') + 1);
+    const participants = await storageEngine.getAllParticipantsDataByStudy(strippedFilename);
+    return participants;
+  };
```
JackWilb, line 141:
```suggestion
  const fetchParticipants = async (snapshotName: string) => {
    const strippedFilename = snapshotName.slice(snapshotName.indexOf('-') + 1);
    return await storageEngine.getAllParticipantsDataByStudy(strippedFilename);
  };
```

JackWilb (approved):

PR #456, opened 2024-08-12
Add ability to define StringOption arrays as string arrays

JackWilb:
No

Allows a simplification of the config, by allowing the options array to be defined using lists of strings instead of list of objects, e.g.:

```
"options": [
    "Bar",
    "Bubble",
    "Pie",
    "Stacked Bar"
]
```

TODOs:
- [x] Update relevant documentation
- [x] Check docs repo for updates (https://github.com/revisit-studies/reVISit-studies.github.io/pull/66)

### DIFF
--- a/public/demo-VLAT-full-randomized/config.json
+++ b/public/demo-VLAT-full-randomized/config.json
@@ -109,28 +109,7 @@
                     "required": true,
                     "location": "sidebar",
                     "type": "radio",
-                    "options": [
-                        {
-                            "label": "$57.36",
-                            "value": "$57.36"
-                        },
-                        {
-                            "label": "$47.82",
-                            "value": "$47.82"
-                        },
-                        {
-                            "label": "$50.24",
-                            "value": "$50.24"
-                        },
-                        {
-                            "label": "$39.72",
-                            "value": "$39.72"
-                        },
-                        {
-                            "label": "Skip",
-                            "value": "Skip"
-                        }
-                    ]
+                    "options": ["57.36", "47.82", "50.24", "39.72", "Skip"]
                 }
             ]
         },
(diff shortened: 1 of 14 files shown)


lesegongwane:
This PR (Pull Request) introduces an enhancement to the data visualization options by adding support for multiple chart types: Bar, Bubble, Pie, and Stacked Bar. The goal is to provide users with a broader set of tools to represent their data in a way that best suits their analytical needs. Here's a breakdown of why each chart type was added and the specific needs they address:

Bar Chart: Bar charts are one of the most commonly used visualization tools for comparing categorical data. They are effective at showing differences in quantities across categories, making them ideal for visualizing discrete data points. By adding support for bar charts, this PR ensures that users can easily compare different groups or categories, which is crucial for data analysis and reporting.

Bubble Chart: Bubble charts are an extension of scatter plots that incorporate a third dimension, represented by the size of the bubbles. This type of visualization is particularly useful for displaying and comparing relationships between three variables. Adding bubble charts addresses the need for more advanced multi-dimensional data visualization, allowing users to explore correlations and trends across multiple variables in a single, visually compelling chart.

Pie Chart: Pie charts are a classic way to represent parts of a whole. They are especially useful for visualizing proportions and percentages. By including pie charts in this update, the PR provides users with a straightforward way to illustrate how different segments contribute to a total, making it easier to communicate the composition of data sets at a glance.

Stacked Bar Chart: Stacked bar charts build on the traditional bar chart by layering multiple data series on top of each other. This allows users to compare not only total values across categories but also the contribution of individual components within those totals. The addition of stacked bar charts addresses the need for more nuanced comparisons, particularly when users want to highlight both overall trends and the breakdown of those trends into constituent parts.

Why This PR is Needed:
This PR is necessary to enhance the flexibility and expressiveness of the data visualization options available to users. By incorporating these chart types, the update addresses the diverse needs of users who require different ways to represent and analyze their data. Whether it's for comparing categorical data, analyzing relationships between multiple variables, understanding proportions, or dissecting aggregated data, these new visualization tools empower users to create more informative and tailored data presentations.

This PR ultimately helps improve the overall user experience by providing a more comprehensive set of visualization tools, enabling users to extract deeper insights from their data with greater ease and precision.

bbollen23 (approved):
You mentioned before about turning the conversion for taking in a string or object and mapping it to the correct label/value, but I don't think that's necessary. It's short enough that repeating it doesn't really make the code bloated.

PR #490, opened 2024-11-04
Add participantTags and add sequence block tags when navigating through the sequence

### LINKED ISSUE #474: Add block tags

### LINKED ISSUE #475: Add block tags to the participant data in download

### LINKED ISSUE #485: Add a marker to the participant data that shows which blocks were entered

JackWilb:
Closes #474
Closes #475 
Closes #485 

This pull request introduces `participantTags`. Participant tags exist to allow tagging whole participants based on some criteria. The initial, automatic tagging that I've added is to tag when a participant enters into a sequence block with an ID. This could be useful if there is branching logic, to quickly find participants that took path A vs path B.

The code changes are as follows:
- Updated `ParticipantData` interface to include `participantTags` property.
- Added methods to get, add, and remove participant tags in the storage engines.
- Updated the ComponentController to fetch the block for the current step and add it as a participant tag automatically.
- Modified the skip logic test to verify the correct handling of participant tags, ensuring tags are added and removed appropriately.

Screenshot 2024-11-04 at 1 41 21 PM

TODOs:
- [x] Update relevant documentation (typedoc)
- [x] Add the tags to the analysis table
- [x] Add the tags to the tidy data download

### DIFF
--- a/src/storage/engines/LocalStorageEngine.ts
+++ b/src/storage/engines/LocalStorageEngine.ts
@@ -249,6 +249,47 @@ export class LocalStorageEngine extends StorageEngine {
     return await this.studyDatabase.getItem(participantId) as ParticipantData | null;
   }
 
+  async getParticipantTags(): Promise<string[]> {
+    if (!this._verifyStudyDatabase(this.studyDatabase)) {
+      throw new Error('Study database not initialized');
+    }
+
+    const participantData = await this.getParticipantData();
+    if (!participantData) {
+      throw new Error('Participant not initialized');
+    }
+
+    return participantData.participantTags;
+  }
+
+  async addParticipantTags(tags: string[]): Promise<void> {
+    if (!this._verifyStudyDatabase(this.studyDatabase)) {
+      throw new Error('Study database not initialized');
+    }
+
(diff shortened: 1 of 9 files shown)


src/storage/engines/LocalStorageEngine.ts
```diff
@@ -248,6 +249,47 @@ export class LocalStorageEngine extends StorageEngine {
...
+
+    const participantData = await this.getParticipantData();
+    if (!participantData) {
+      throw new Error('Participant not initialized');
+    }
+
+    participantData.participantTags = participantData.participantTags.filter((tag) => !tags.includes(tag));
+    await this.studyDatabase.setItem(this.currentParticipantId as string, participantData);
+  }
+
```
bbollen23, line 292: Just a general comment for future: Is there a way to automatically wrap these functions in something that checks for participant data and study database being initialized? I'm not sure how we would go about this, but it seems tedious to have to write this over and over again. At the very least, maybe we should just create a separate function that abstracts this.

bbollen23 (approved):
Just one comment I made. Doesn't need to be handled right now.

PR #509, opened 2024-12-13
Provenance + audio replay

ZachCutler04:

Adds in replay functionality, accessible via the provenance tab in the analysis view. Replay should work with both/one of audio and provenance data, though audio will only work via firebase (provenance will work for both)

Main todo is that we don't check for past versions of config currently, so old configs may run into problems during replay.


src/components/interface/AnalysisFooter.tsx
```diff
@@ -0,0 +1,119 @@
...
+  const { value: participant, status: loadingPartStatus } = useAsync(getParticipantData, [participantId, storageEngine]);
+  const { value: allParticipants } = useAsync(getAllParticipantIds, [storageEngine]);
+
+  const nextParticipantNameAndIndex: [string, number] = useMemo(() => {
+    if (allParticipants && participant && participantId && currentComponent) {
+      const filteredParticipants = allParticipants.filter((part) => part.participantConfigHash === participant.participantConfigHash);
+      const index = filteredParticipants.findIndex((part) => part.participantId === participantId);
+      const nextPart = index < filteredParticipants.length - 1 ? filteredParticipants[index + 1] : filteredParticipants[0];
+
+      return [nextPart.participantId, getSequenceFlatMap(nextPart.sequence).indexOf(currentComponent)];
```
JackWilb, line 62: This doesn't work for repeated components

JackWilb (approved):

JackWilb (approved):
Notes of things to fix:
- Need to disable reviewer steps panel in replay
- Update play pause button
- Design pass on the replay timeline visualization
-

PR #595, opened 2025-01-28
Fix issue with slider label alignment, hide not publicly accessible studies from "your studies" on landing page

JackWilb:
No, user reported bugs

Slider labels could overflow, this should fix that with a small CSS tweak.

We want to be able to hide studies in the "your studies" tab on the main page. If the revisit mode has `analyticsInterfacePubliclyAccessible = true`, then it renders, otherwise, it does not render on the front page. Obviously if the user is an admin, it still shows.

Slider:
Screenshot 2025-01-27 at 6 14 44 PM

No

### DIFF
--- a/src/components/response/SliderInput.tsx
+++ b/src/components/response/SliderInput.tsx
@@ -7,6 +7,7 @@ import { useMemo } from 'react';
 import { SliderResponse } from '../../parser/types';
 import { generateErrorMessage } from './utils';
 import { ReactMarkdownWrapper } from '../ReactMarkdownWrapper';
+import classes from './SliderInput.module.css';
 
 export function SliderInput({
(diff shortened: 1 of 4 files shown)


src/components/response/SliderInput.tsx
```diff
@@ -53,6 +54,7 @@ export function SliderInput({
         {...answer}
         h={40}
         mt={4}
+        classNames={{ markLabel: classes.markLabel }}
```
bbollen23, line 57: Can just do this since you only have markLabel in the module anyway. But yours is also a bit more clear

```suggestion
        classNames={classes}
```

bbollen23 (approved):
Just one suggested change, looks good

PR #600, opened 2025-02-03
Adding custom function logic to the config

### LINKED ISSUE #592: Add logic for step study design

ZachCutler04:
Closes #592 

Adds functionality to allow users to pass in custom functions to determine order of components, instead of fixed, random, or latinSquare. 

Accessible via the config sequence, instead of a typical block users can pass in a 'dynamic' block, which looks something like 

```json 
        {
            "id": "loopingBlock",
            "order": "dynamic",
            "func": "test-step-logic/func.ts"
        },
```

Which also can accept an optional parameters which gets passed to the function. The function must return a component name which will be the next task, or null, to leave the block. The function also optionally can return a parameters object, which will override the components parameters.


src/store/store.tsx
```diff
@@ -174,13 +208,7 @@ export async function studyStoreCreator(
...
       ) => {
         if (!state.trialValidation[payload.identifier]) {
-          state.trialValidation[payload.identifier] = {
-            aboveStimulus: { valid: false, values: {} },
-            belowStimulus: { valid: false, values: {} },
-            sidebar: { valid: false, values: {} },
-            stimulus: { valid: false, values: {} },
-            provenanceGraph: undefined,
-          };
+          return;
```
JackWilb, line 211: How does this not bug?

PR #649, opened 2025-02-28
Add provenance tracking and replay to form elements

### LINKED ISSUE #598: Track response progress in forms
We currently don't track response progress in forms via provenance tracking. It might be interesting in some cases though, to see how ppl change values e.g., on sliders, and how they type text. 

Basically, we want to provenance track every single input event. A key press, a drag, a release, etc. All of it time-stamped.

JackWilb:
Closes #598 

Adds trrack to track the mantine `useForm` variable. We modified the response block each have a track instance and we now have up to 4 trrack instances. We coordinate the rehydration at the response block and stimulus level so code is not duplicated.

TODO

TODOs:
- [ ] Add the rehydration


JackWilb:
2 notes:
- One bug, it doesn't clear if there is no provenance on the next task
- We need to test with non-linear stimulus history. We are implicitly linear for the form elements

PR #703, opened 2025-04-02
Add dynamic blocks demo

yeonkim1213:
Add dynamic blocks demo


src/public/demo-dynamic/assets/dynamic.tsx
JackWilb, line 1: Please add some more inline comments to show why doing each logic step in the function so someone could read this and understand it without looking at the specifics of the code

JackWilb (approved):

PR #716, opened 2025-04-17
Minimum screen size warning

### LINKED ISSUE #633: Enforce device check
Desktop, screen resolution

yeonkim1213:
Closes #633 

To inform minimum screen size to the participant if their screen is too small.
Study designers can specify minWidthSize and minHeightSize to uiConfig in config.json

It rejects participants if not resized in 1 minute (It will not reject if the participant resizes)
It will show the training failed page after 1 minute.

#### 1 minute timer

https://github.com/user-attachments/assets/d9060e42-fe27-4003-abad-60de70be89f9

#### After 1 minute

rejected

#### Resize

https://github.com/user-attachments/assets/6948f254-0d2d-4193-9f24-c1e42aa3ece1

#### Different screen size
**Desktop**

desktop1

desktop2

**Mobile**

iPhone SE
iphoneSE

Pixel 7
pixel7

Galaxy S8+
galaxys8+

#### Rejected participant  

image

TODOs:
- [x] Timeout
- [x] Reject current participant
- [x] Add 1-minute timer to the warning
- [x] Edit displayed message


src/parser/types.ts
```diff
@@ -121,6 +121,10 @@ export interface UIConfig {
   participantNameField?: string;
   /** Whether enter key should move to the next question. Defaults to false. */
   nextOnEnter?: boolean;
+  /** The minimum screen width size for the study */
+  minWidthSize?: number;
+  /** The minimum screen height size for the study */
+  minHeightSize?: number;
```
JackWilb, line 127: I'd like to see a few more types here. We need a way to force rejection for participants whose screen size is not within the range specified here. We should also allow for max width/height (for in the future when we want to force mobile)

src/components/interface/ResolutionWarning.tsx
```diff
@@ -0,0 +1,63 @@
...
+          {minHeight}
+          ).
+        </Text>
+        <Text size="md" ta="center">
+          Some content may not display correctly. Please use a device with a larger screen or resize your browser window.
+        </Text>
+        <Group mt="lg">
+          <Button color="yellow" onClick={() => { setShowWarning(false); }} variant="outline">
+            Continue anyway
+          </Button>
```
JackWilb, line 55: This should be controlled by a new type, which I described in a comment on the types file

src/components/interface/ResolutionWarning.tsx
```diff
@@ -0,0 +1,63 @@
...
+          <Button onClick={() => navigate('/')} variant="filled">
+            Go to main page
+          </Button>
```
JackWilb, line 58: We don't want people to navigate to /. Maybe remove this button

src/components/interface/ResolutionWarning.tsx
```diff
@@ -0,0 +1,63 @@
...
+      if (window.innerWidth < minWidth || window.innerHeight < minHeight) {
+        setShowWarning(true);
+      } else {
+        setShowWarning(false);
+      }
+    };
+    handleResize();
+    window.addEventListener('resize', handleResize);
+    return () => window.removeEventListener('resize', handleResize);
+  }, [minWidth, minHeight]);
```
JackWilb, line 29: Good job handling resize!

src/components/interface/ResolutionWarning.tsx
```diff
@@ -0,0 +1,63 @@
...
+import { IconAlertTriangle } from '@tabler/icons-react';
+import { useNavigate } from 'react-router';
+import { useStoreSelector } from '../../store/store';
+
+export function ResolutionWarning() {
+  const navigate = useNavigate();
+  const studyConfig = useStoreSelector((state) => state.config);
+
+  const minWidth = studyConfig.uiConfig.minWidthSize!;
+  const minHeight = studyConfig.uiConfig.minHeightSize!;
```
JackWilb, line 14: You're forcing typescript to think these are defined, but maybe it's better for them to possibly be undefined? You'll have to update the logic in the resize handler, though
```suggestion
  const minWidth = studyConfig.uiConfig.minWidthSize;
  const minHeight = studyConfig.uiConfig.minHeightSize;
```

src/components/interface/ResolutionWarning.tsx
```diff
@@ -0,0 +1,63 @@
...
+    window.addEventListener('resize', handleResize);
+    return () => window.removeEventListener('resize', handleResize);
+  }, [minWidth, minHeight]);
+
+  if (!showWarning) {
+    return null;
+  }
+
+  return (
+    <Modal opened withCloseButton={false} onClose={() => { }} fullScreen>
```
JackWilb, line 36: Is `onClose` required or can it be removed?

JackWilb, line 36: Have you tried this modal on small screen sizes? Do we get good looking full screen support here? Please attach some more pictures of different screen sizes

JackWilb (changes requested):

src/components/interface/ResolutionWarning.tsx
yeonkim1213: Has been removed.

src/components/interface/ResolutionWarning.tsx
yeonkim1213: Removed the button.

src/parser/types.ts
yeonkim1213: During the ReVISit meeting, we decided to use only minWidthSize and minHeightSize.

src/components/interface/ResolutionWarning.tsx
yeonkim1213: Decided to remove this button (from the ReVISit meeting)

src/components/interface/ResolutionWarning.tsx
yeonkim1213: Looks good on small screen sizes as well - I've attached more pictures (desktop ver., mobile ver.)

yeonkim1213: Yes, onClose is required, but because we don’t want users to close the modal, I left it blank

yeonkim1213:
Tried forcing the window to close after 5 seconds when the screen is too small, but it seems that browser permissions don’t allow it

src/components/interface/ResolutionWarning.tsx
```diff
@@ -0,0 +1,124 @@
...
+} from '@mantine/core';
+import { useEffect, useState, useRef } from 'react';
+import { IconAlertTriangle } from '@tabler/icons-react';
+import { useNavigate } from 'react-router';
+import { useStoreSelector } from '../../store/store';
+import { useStorageEngine } from '../../storage/storageEngineHooks';
+import { useStudyId } from '../../routes/utils';
+
+export function ResolutionWarning() {
+  const studyConfig = useStoreSelector((state) => state.config);
```
JackWilb, line 12: Better to use our hook
```suggestion
  const studyConfig = useStudyConfig();
```

JackWilb (approved):

PR #724, opened 2025-04-22
Dynamic blocks - question harder, add checkAnswerOnEnter

### LINKED ISSUE #723: Add checkAnswerOnEnter
In the dynamic blocks example, keyboard‑only navigation is not supported since I am checking the answer before going to the next page. 
Add checkAnswerOnEnter to enable keyboard navigation.

yeonkim1213:
Closes #723 

1. Made the dynamic blocks questions harder
- Changed initial saturation value from Left: 0%, Right: 100% to Left: 30%, Right: 70%
- Increased the saturation adjustment value  from 5% to 10% for more noticeable difficulty changes
- Now compares the actual HSL values of the left and right squares to determine the correct answer

2. Added 'Same' option
- The saturation value of left and right squares will eventually be the same as user goes through the question
- If user clicks on the 'Same' option and the answer is correct, it will exit dynamic blocks since it won't make sense to make the questions harder. In default, user will have to go through 10 questions.

3. Addchecking answer on enter while `nextOnEnter = true`
- Users can navigate through questions with keyboard now

#### Using mouse

https://github.com/user-attachments/assets/06fe2d9b-08ef-4102-93b5-f7b6f4c49370

#### Using keyboard

https://github.com/user-attachments/assets/a0d32d7e-db65-439f-b30e-f214cafb3b09

TODOs:

- [x] Update test
- [x] Remove checkAnswerOnEnter and move codes to nextOnEnter

### DIFF
--- a/src/components/response/ResponseBlock.tsx
+++ b/src/components/response/ResponseBlock.tsx
@@ -261,7 +261,7 @@ export function ResponseBlock({
       }
     };
 
-    if (studyConfig.uiConfig.checkAnswerOnEnter) {
+    if (studyConfig.uiConfig.nextOnEnter) {
       window.addEventListener('keydown', handleKeyDown);
       return () => {
         window.removeEventListener('keydown', handleKeyDown);
--- a/src/components/response/ResponseBlock.tsx
+++ b/src/components/response/ResponseBlock.tsx
@@ -255,13 +255,13 @@ export function ResponseBlock({
   }, [attemptsUsed, responsesWithDefaults, configInUse, hasCorrectAnswerFeedback, trainingAttempts, allowFailedTraining, storageEngine, studyId, navigate, identifier, storeDispatch, answerValidator, alertConfig, saveIncorrectAnswer]);
 
   useEffect(() => {
-    const handleKeyDown = (event: KeyboardEvent) => {
-      if (event.key === 'Enter') {
-        checkAnswerProvideFeedback();
-      }
-    };
-
     if (studyConfig.uiConfig.nextOnEnter) {
+      const handleKeyDown = (event: KeyboardEvent) => {
+        if (event.key === 'Enter') {
+          checkAnswerProvideFeedback();
+        }
+      };
+
       window.addEventListener('keydown', handleKeyDown);
       return () => {
         window.removeEventListener('keydown', handleKeyDown);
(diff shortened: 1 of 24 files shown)


src/components/response/ResponseBlock.tsx
```diff
@@ -243,7 +252,23 @@ export function ResponseBlock({
...
+    };
+
+    if (studyConfig.uiConfig.nextOnEnter) {
+      window.addEventListener('keydown', handleKeyDown);
+      return () => {
+        window.removeEventListener('keydown', handleKeyDown);
+      };
+    }
+    return () => {};
+  }, [checkAnswerProvideFeedback, studyConfig]);
```
JackWilb, line 271: This appears to always be active. Do we only want to activate this with `nextOnEnter`?

src/public/demo-dynamic/assets/dynamic.tsx
```diff
@@ -14,39 +14,50 @@ export default function dynamic({ answers }: JumpFunctionParameters<never>): Jum
...
 
   let message = 'The answer difficulty will change based on your last answer';
   let color = 'blue';
 
-  if (lastAnswer.answer.buttonResponse === 'Right') {
+  if (lastAnswer) {
+    const isCorrect = (lastAnswer.answer.buttonResponse === 'Left' && lastAnswer.parameters.left > lastAnswer.parameters.right)
+                      || (lastAnswer.answer.buttonResponse === 'Right' && lastAnswer.parameters.right > lastAnswer.parameters.left)
+                      || (lastAnswer.answer.buttonResponse === 'Same' && lastAnswer.parameters.left === lastAnswer.parameters.right);
+
```
JackWilb, line 28: Can we move this up outside the if statement so we don't have to redefine this multiple times?

JackWilb (changes requested):

src/public/demo-dynamic/assets/dynamic.tsx
yeonkim1213: Moved it outside of the if statement

src/components/response/ResponseBlock.tsx
yeonkim1213: moved if statement to set up the event listener only when nextOnEnter is true

PR #734, opened 2025-05-09
Randomize form elements

### LINKED ISSUE #433: Randomization of responses
Currently we have randomization of stimuli.
But we don't have randomization of responses array within each stimuli.

### LINKED ISSUE #718: randomize form elements
It should be possible to randomize all form elements that are on a page. 
This randomization should also be indicated in the side bar.

yeonkim1213:
Closes #718
Closes #433

Allow the study designer to randomize all form elements on the page.

#### Randomize Form

https://github.com/user-attachments/assets/fdae6b12-76c8-43d5-aca2-2b62cd219f91

#### Replay
Replay is shown in same order that study participant saw

https://github.com/user-attachments/assets/40aa704d-3e09-4c19-a9dd-efb915da39e8

#### Die icon on sidebar and 'random' message shown if hovered over

https://github.com/user-attachments/assets/b27234ad-75a7-4df4-a0b2-253b35e8063d

TODOs:
- [x] Check DB is updated correctly
- [x] Check if order is consistent for the same user


src/utils/handleResponseRandomization.ts
```diff
@@ -1,5 +1,18 @@
...
+    if (componentConfig.randomizeForm) {
+      const shuffled = [...componentConfig.response]
+        .map((value) => ({ value, sort: Math.random() }))
+        .sort((a, b) => a.sort - b.sort)
+        .map(({ value }) => value);
+      acc[response.id] = [shuffled.findIndex((value) => value.id === response.id)];
+    }
+    return acc;
+  }, {} as Record<string, number[]>);
+}
```
JackWilb, line 14: This looks buggy. For each item in the response (each question) we produce a new shuffle and then find the index in the new shuffle. I don't think that's quite what we want.

I think you only need this
```
const shuffled = [...componentConfig.response]
        .map((value) => ({ value, sort: Math.random() }))
        .sort((a, b) => a.sort - b.sort)
        .map(({ value }) => value);
```

src/components/response/ResponseBlock.tsx
```diff
@@ -51,18 +52,34 @@ export function ResponseBlock({
...
+  const formOrders : Record<string, number[]> = useMemo(() => storedAnswerData?.formOrder || {}, [storedAnswerData]);
 
   const studyId = useStudyId();
 
   const navigate = useNavigate();
 
   const configInUse = config as IndividualComponent;
 
-  const responses = useMemo(() => configInUse?.response?.filter((r) => (r.location ? r.location === location : location === 'belowStimulus')) || [], [configInUse?.response, location]);
+  const filteredResponses = useMemo(() => configInUse?.response?.filter((r) => (r.location ? r.location === location : location === 'belowStimulus')) || [], [configInUse?.response, location]);
```
JackWilb, line 69: Can we not just use the formOrder here and skip the memo below?

```suggestion
  const filteredResponses = useMemo(() => formOrders.filter((r) => (r.location ? r.location === location : location === 'belowStimulus')) || [], [configInUse?.response, location]);
```

src/components/response/ResponseBlock.tsx
```diff
@@ -51,18 +52,34 @@ export function ResponseBlock({
...
+    if (formOrders) {
+      // Sort responses based on their order in formOrders
+      return [...filteredResponses]
+        .map((value) => ({ value, sort: formOrders[value.id]?.[0] }))
+        .sort((a, b) => a.sort - b.sort)
+        .map(({ value }) => value);
+    }
+    // Fixed order
+    return filteredResponses;
+  }, [filteredResponses, formOrders]);
```
JackWilb, line 82: If my fix above works, this can be removed

yeonkim1213: Removed

PR #763, opened 2025-07-01
Allow variable injection at the study end

### LINKED ISSUE #739: Allow variable injection in the study end message
Some recruitment platforms require an id to be passed back to them to verify completion. This needs to be added on to the URL, but we currently have no way to do that.

There should be some way to specify a variable replacement that injects it into the end message link. This could require some custom syntax in the config, similar to format strings.

**Additional context**
Requested by Jeremy Block

yeonkim1213:
Closes #739 

Allow variable injection at the study end

Uses
``` ts
/** The message to display when the study ends. Supports templating with the participant ID. e.g. "Thank you for completing the study. You may click this link and return to Prolific: [Go to Prolific](https://app.prolific.com/submissions/complete?cc=StudyID&PROLIFIC_ID={PROLIFIC_ID}))" */
  studyEndMsg?: string;
/** If the participant ID is passed in the URL, this is the name of the querystring parameter that is used to capture the participant ID (e.g. PROLIFIC_ID). This will allow a user to continue a study on different devices and browsers. */
  urlParticipantIdParam?: string;
```

```json
"uiConfig": {
  "studyEndMsg": "Thank you for completing the study. You may click this link and return to Prolific: [https://app.prolific.com/submissions/complete?cc=CE96X7ST&PROLIFIC_ID={PROLIFIC_ID}](https://app.prolific.com/submissions/complete?cc=CE96X7ST&PROLIFIC_ID={PROLIFIC_ID})",
  "urlParticipantIdParam": "PROLIFIC_PID",
}
```
e.g. http://revisit.dev/example-brush-interactions/SURHUmJIblo3Uk4xUUhNUHQ1VTRIQT09?PROLIFIC_PID=123456

image

image

TODOs:
- [x] Update studies to use studyEndMsg template
- [x] Add example url template to comment

### DIFF
--- a/src/components/StudyEnd.tsx
+++ b/src/components/StudyEnd.tsx
@@ -14,6 +14,7 @@ import { download } from './downloader/DownloadTidy';
 import { useStudyId } from '../routes/utils';
 import { useIsAnalysis } from '../store/hooks/useIsAnalysis';
 import type { Response } from '../parser/types';
+import { studyComponentToIndividualComponent } from '../utils/handleComponentInheritance';
 
 export function StudyEnd() {
   const studyConfig = useStudyConfig();
@@ -115,11 +116,8 @@ export function StudyEnd() {
     const templateData: Record<string, string> = { PARTICIPANT_ID: participantId };
 
     const responses = Object.values(studyConfig.components)
-      .flatMap((component) => (
-        Array.isArray((component as { response?: Response[] }).response)
-          ? (component as { response: Response[] }).response
-          : []
-      ));
+      .map((component) => studyComponentToIndividualComponent(component, studyConfig))
+      .flatMap((component) => (Array.isArray(component.response) ? component.response : []));
 
     const fieldId = responses.find((r: Response) => r.paramCapture === urlParticipantIdParam)?.id;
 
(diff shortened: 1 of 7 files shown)


src/components/StudyEnd.tsx
```diff
@@ -100,12 +104,42 @@ export function StudyEnd() {
...
+    }
+
+    const templateData: Record<string, string> = { PARTICIPANT_ID: participantId };
+
+    const responses = Object.values(studyConfig.components)
+      .flatMap((component) => (
+        Array.isArray((component as { response?: Response[] }).response)
+          ? (component as { response: Response[] }).response
+          : []
+      ));
```
JackWilb, line 122: The reason this was likely complaining about types is because of base components. Using the type you defined here adds a bug with base components, because responses are not always guaranteed to be available at the `components` level (e.g. they are defined on the base component instead). You will need to use the function that takes a component and fully specifies it. That will also let you remove this type cast.

src/components/StudyEnd.tsx
```diff
@@ -100,12 +104,42 @@ export function StudyEnd() {
...
 
+  const processedStudyEndMsg = useMemo(() => {
+    const { studyEndMsg } = studyConfig.uiConfig;
+    const { urlParticipantIdParam } = studyConfig.uiConfig;
+
+    if (!urlParticipantIdParam || !studyEndMsg?.includes('{')) {
+      return studyEndMsg;
+    }
+
+    const templateData: Record<string, string> = { PARTICIPANT_ID: participantId };
```
JackWilb, line 115: How would we add other values to this?

PR #773, opened 2025-07-10
Enable YAML configurations

### LINKED ISSUE #748: Spec Format Middleware? (support yaml, etc)
A request we've gotten (at least once) is to support other formats, such as yaml and friends. While I think our choice to center JSON is a reasonable mediation of factors (maintainability, schema, start up costs, etc), I wonder if it might be useful to make it so that users can optionally use other formats if they like. This would allow users to work in whatever serialization format they like. 

Adding support for other formats obviously adds maintainability surface, which for a features of this type is not necessarily ideal. As a solution I suggest we might develop a notion of middle ware that enables straight-forwardish translation of specs into our target format. That way, if a user wants to use something exotic (like dhall or hjson) they can just provide a translation function that takes in whatever and returns a valid spec. Such a solution also opens opportunities for other middleware tasks, perhaps such as macros or language extensions.

JackWilb:
Closes #748

This adds a YAML parser to the app and allows configs to be written and parsed from YAML. I also added a settings.json file that configures the vscode YAML parser to use the on disk schema so that the YAML can be validated and autocompleted in editor. I think this is a fine extension for now, and if we get more requests for other formats, we can re-engineer to support more.

See deployment for demo-yaml

TODOs:

### DIFF
--- a/src/parser/parser.ts
+++ b/src/parser/parser.ts
@@ -1,4 +1,5 @@
 import Ajv from 'ajv';
+import { parseDocument } from 'yaml';
 import configSchema from './StudyConfigSchema.json';
 import globalSchema from './GlobalConfigSchema.json';
 import {
@@ -187,10 +188,18 @@ export async function parseStudyConfig(fileData: string): Promise<ParsedConfig<S
   let data: StudyConfig | undefined;
 
   try {
+    // Try JSON parse first
     data = JSON.parse(fileData);
     validatedData = studyValidate(data) as boolean;
   } catch {
-    validatedData = false;
+    // Try yaml parse
+    try {
+      data = parseDocument(fileData).toJSON() as StudyConfig;
+      validatedData = studyValidate(data) as boolean;
+    } catch (e) {
+      console.error('Error parsing study config file:', e);
+      validatedData = false;
+    }
   }
 
   let errors: Required<ParsedConfig<StudyConfig>>['errors'] = studyValidate.errors || [];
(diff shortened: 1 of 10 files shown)


mcnuttandrew:
First off it looks nice! Thanks for putting this together! I have mixed feelings about this!

On the positive side
+ This is a pretty small change and gives us a nice bullet point in the next release blog post
+ Seems like it address the full scope of the additional formats people have asked for

On the negative side
- It means we bundle another bunch of things in this already gigantic code base (and therein have to also look after versioning of those deps)
- While we have removed support for a format before (hjson -> json), I worry that making this kind of foundational input change won't be as easy in the future. This is somewhat in part bc we don't have a good way to tell what users are using what features (perhaps this suggests some sort of scrape level audit at some point?). If we decide down the road to drop support for yaml, communicating that to users may be tedious/annoying. 

More minor: would be cool to just modify one of the existing demos to use yaml rather than having yet another one (yaml pun). The actual scope of changes is pretty small, so swapping a extant json version to a yaml version seems useful

src/parser/parser.ts
yeonkim1213, line 1: Can we add `as StudyConfig` to the JSON parse as well for explicit type assertions?
`data = JSON.parse(fileData); as StudyConfig`

JackWilb: This is not needed because of the type, set above. The yaml needs to be coerced, because the types are less useful from this library

PR #786, opened 2025-08-04
Update style

yeonkim1213:
- Update demo-style styling
- Added `minHeight` `minWidth` `maxWidth` `maxHeight`
 
#### demo-style

- Set `border: 0` to iframe (website) components to remove default borders
- Updated google form styled page from `width: 80%` to `width: 50%`

image

TODOs:
None


public/demo-style/assets/style/form.css
```diff
@@ -20,16 +20,15 @@ body {
...
 
 .response {
   background-color: #ffffff;
-  margin: 15px auto 0;
+  margin: 10px auto 0;
   padding: 20px;
   border: 2px solid #e3e8ed;
   border-radius: 12px;
-  width: 80%;
+  width: 70%;
```
JackWilb, line 31: Let's go a little smaller, 60/50 so the change is more visible for Alex

public/demo-style/assets/style/layout.css
```diff
@@ -4,10 +4,14 @@
   gap: 20px;
   margin-bottom: 20px;
 }
+
 #layout-title {
   grid-row: 1;
   grid-column: span 6;
   text-align: center;
+  border: none;
```
JackWilb, line 12: If this is for the iframe we could remove.

I find these CSS files pretty inscrutable in general. Can we add some comments across all these files so they're more usable for people copying the structure? The comments should say what it updates by component name and response name (if applicable)

src/components/response/ButtonsInput.tsx
```diff
@@ -54,7 +54,7 @@ export function ButtonsInput({
         error={error}
         style={{ '--input-description-size': 'calc(var(--mantine-font-size-md) - calc(0.125rem * var(--mantine-scale)))' }}
       >
-        <Flex justify="space-between" align="center" gap="xl" mt="xs">
+        <Flex justify="space-between" align="center" gap="xl" mt="xs" style={{ overflow: 'hidden' }}>
```
JackWilb, line 57: Why is this needed? Seems like this could hide the overflowing radio options and make it impossible to select some of them

src/controllers/IframeController.tsx
```diff
@@ -17,7 +17,7 @@ export function IframeController({ currentConfig, provState, answers }: { curren
   const dispatch = useDispatch();
   const identifier = useCurrentIdentifier();
   const [height, setHeight] = useState(800);
-  const iframeStyle = useMemo(() => ({ height: `${height}px`, width: '100%' }), [height]);
+  const iframeStyle = useMemo(() => ({ height: `${height}px`, width: '100%', border: 0 }), [height]);
```
JackWilb, line 20: This will conflict with my changes. Let's be careful about merge order and making sure we don't lose this in the conflicts

public/libraries/calvi/config.json
JackWilb, line 1: I like the simplification, but instead of moving this to a CSS file, could we use a base component instead to reduce the redundant style blocks?

JackWilb (changes requested):

public/demo-html-input/config.json
```diff
@@ -35,7 +35,8 @@
         }
       ],
       "path": "demo-html-input/assets/bar-chart-interaction.html",
-      "instructionLocation": "aboveStimulus"
+      "instructionLocation": "aboveStimulus",
+      "stylesheetPath": "demo-html-input/assets/barChart.css"
```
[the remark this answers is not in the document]
yeonkim1213: Removed

public/demo-html-input/assets/barChart.css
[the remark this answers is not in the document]
yeonkim1213: Not required anymore - removed the file

public/demo-style/assets/style/form.css
yeonkim1213: Updated to `width: 50%`

public/demo-style/assets/style/layout.css
yeonkim1213: Added comments in the CSS.

`border: none` is for markdown - I thought having no border around the title would look better.

image

src/components/response/ButtonsInput.tsx
yeonkim1213: Reverted - I added that because the buttons looked overflowed on the smaller screen size but I definetly agree that this could interfere with selecting radio options

PR #809, opened 2025-08-26
Custom audio waveform to improve performance

### LINKED ISSUE #684: Fix audio waveform for Firefox
It's currently missing because of a bug with wavesurfer

hilsonshrestha:
Closes #684 

Addresses the high CPU usage caused by Wavesurfer when rendering even small waveforms in the app header during think-aloud sessions.

This PR can possibly be extended to
- Accepting an audio stream (e.g., from screen recording or another component) as a prop.
- Storing waveform data during recording, reducing the need to repeatedly decode the entire audio/video file during analysis.

image

### DIFF
--- a/src/components/interface/RecordingAudioWaveform.tsx
+++ b/src/components/interface/RecordingAudioWaveform.tsx
@@ -1,15 +1,6 @@
 import { Flex } from '@mantine/core';
 import { useRef, useEffect, useState } from 'react';
 
-interface AudioVisualizerProps {
-  width?: number;
-  height?: number;
-  fps?: number;
-  fftSize?: 32 | 64 | 128 | 256 | 512 | 1024 | 2048 | 4096 | 8192 | 16384;
-  barColor?: string;
-  barWidth?: number;
-}
-
 export function RecordingAudioWaveform({
   width = 60,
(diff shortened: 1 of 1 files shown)


src/components/interface/RecordingAudioWaveform.tsx
```diff
@@ -1,52 +1,148 @@
...
+}
+
+export function RecordingAudioWaveform({
+  width = 60,
+  height = 36,
+  fps = 30,
+  fftSize = 256,
+  barColor = '#FA5252',
+  barWidth = 2,
+}: AudioVisualizerProps) {
```
JackWilb, line 20:
```suggestion
export function RecordingAudioWaveform({
  width = 60,
  height = 36,
  fps = 30,
  fftSize = 256,
  barColor = '#FA5252',
  barWidth = 2,
}: {
  width?: number;
  height?: number;
  fps?: number;
  fftSize?: 32 | 64 | 128 | 256 | 512 | 1024 | 2048 | 4096 | 8192 | 16384;
  barColor?: string;
  barWidth?: number;
}) {
```

JackWilb (approved):

PR #915, opened 2025-10-15
Virtual Chinrest Plugin 

mika-long:
No. 

This PR creates a new virtual chin-rest plugin for the library of plugins. It reimplements the original code of Qisheng Li (https://github.com/QishengLi/virtual_chinrest) for calculating viewing distances in large-scale, psychophysical online studies. 

The "calibration process" is consisted of two parts. The first elicits the pixel to physical distance from participants to account for differences in screen sizes and resolutions. 

Screenshot 2025-10-15 at 6 37 16 PM

The second part uses the fact that the human blind spot is 13.5 angles and uses a simple experiment to backward deduce the distance from the eye to the screen. 

Screenshot 2025-10-15 at 6 39 26 PM

Unclear, not that I know of.


mika-long:
Good points! 
* You're right, it does not need to be centered, you can remove the container. (On a related note, is there like a device check we can enforce? Like a participant definitely shouldn't be doing this test on their phones because the width of a credit card is wider than the width of a phone) 
* Yes! For both these tests the sidebar should be disabled.

PR #917, opened 2025-10-16
Stage Label Management

### LINKED ISSUE #874: Label Study Stages
It's often useful to label study stages, such as "debugging" - "pre-pilot" - "pilot" - "main-study" and have these fields available in the dataset. 

We will create a "default" stage. 

Stages should have: 

* a label (that's the only thing that's stored in the data) 
* a color tag (so that it's easy to identify in the participant view)
* a long description
* a number that specifies "target # of participants"

Exactly one stage needs to be active, and we need the UI to activate a stage. When a new stage is created, it should be activated by default. 

We should allow experimenters to create arbitrary stages in the "Manage" tab. 

This could be visuallized in a table; and the tag selection / renaming / coloring UI could be reused.

dyr429:
Closes #874

- Add stage management in the manage tab
- Users can add a stage, give the stage a color, and modifythe  existing stage color
- Add a stage selection. The data displayed in each tab, and the downloaded data, will be filtered by the selected stage.

https://github.com/user-attachments/assets/f00053fc-bef6-4af7-84b5-2a9e384c1d09

### DIFF
--- a/src/analysis/individualStudy/management/StageManagementAccordionItem.tsx
+++ b/src/analysis/individualStudy/management/StageManagementAccordionItem.tsx
@@ -91,25 +91,25 @@ export function StageManagementAccordionItem({ studyId }: { studyId: string }) {
       return;
     }
 
-    // Convert to lowercase
-    const normalizedStageName = newStageName.trim().toLowerCase();
+    const normalizedStageName = newStageName.trim();
+    const lowerCaseName = normalizedStageName.toLowerCase();
 
-    if (normalizedStageName === 'n/a') {
+    if (lowerCaseName === 'n/a') {
       setNewStageError('Stage name "N/A" is reserved and cannot be used');
       return;
     }
 
-    if (normalizedStageName === 'all') {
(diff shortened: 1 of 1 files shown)


src/analysis/individualStudy/management/StageManagementAccordionItem.tsx
```diff
@@ -0,0 +1,291 @@
...
+    if (allStages.some((s) => s.stageName.toLowerCase() === normalizedStageName)) {
+      setNewStageError('A stage with this name already exists');
+      return;
+    }
+
+    setNewStageError('');
+
+    if (storageEngine) {
+      // Add the new stage by setting it as current (which adds it to allStages)
+      await storageEngine.setCurrentStage(studyId, normalizedStageName, newStageColor);
```
JackWilb: Should we consider allowing names with some uppercase characters

JackWilb (commented):
One small thing. We should also take a pass on the storage engine to simplify

PR #941, opened 2025-10-29
Exclude form elements from randomization

### LINKED ISSUE #899: Exclude form elements from randomization
We need to exclude some elements from randomization of questions, otherwise funny things happen, like this here: 

Image

Note the title of the form at the bottom. 

this should probably be a field: 

"exclude-from-randomization", which defaults to false. 

All excluded elements are rendered first in the order they appear. All random elements are rendered next in random order.

jaykim1213:
Closes #899

Exclude form elements from randomization

```ts
export interface BaseResponse {
  /** Exclude response from randomization. If present, will override the responseOrder randomization setting in the components. Defaults to false. */
  excludeFromRandomization?: boolean;
}
```

image

TODOs:
- [ ] Update documentation
- [x] Update `demo-survey`

### DIFF
--- a/src/utils/handleResponseRandomization.ts
+++ b/src/utils/handleResponseRandomization.ts
@@ -4,12 +4,14 @@ export function randomizeForm(componentConfig: IndividualComponent) {
   const response = componentConfig.response.map((r) => r.id);
 
   if (componentConfig.responseOrder === 'random') {
-    return {
-      response: [...response]
-        .map((value) => ({ value, sort: Math.random() }))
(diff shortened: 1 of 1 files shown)


JackWilb (commented):
Seems good, but I would expect that if there was a fixed element, we would want to randomize in 2 groups, the group above the fixed and the group below the fixed. Let's check in the meeting today if that's everyone's understanding

JackWilb (approved):
Approving based on feedback in the meeting. Please make sure to document the above vs below stimulus nuance for randomizing in 2 groups

PR #967, opened 2025-11-19
Add helper function to clean up old modes

jaykim1213:
No

In **#916**, we updated the names of two modes. This PR adds a helper function to update any old mode names to the new ones.

Old version

```ts
export type REVISIT_MODE = 'dataCollectionEnabled' | 'studyNavigatorEnabled' | 'analyticsInterfacePubliclyAccessible';
```

New version

```ts
export type REVISIT_MODE = 'dataCollectionEnabled' | 'developmentModeEnabled' | 'dataSharingEnabled';
```

N/A

- [x] Move cleanup modes to the `storageEngine/types.ts`

### DIFF
--- a/src/storage/engines/FirebaseStorageEngine.ts
+++ b/src/storage/engines/FirebaseStorageEngine.ts
@@ -373,9 +373,15 @@ export class FirebaseStorageEngine extends CloudStorageEngine {
 
     if (revisitModesData.exists()) {
       const modes = revisitModesData.data() as Record<string, boolean>;
-      const cleanedModes = cleanupModes(modes);
-      await setDoc(revisitModesDoc, cleanedModes);
-      return cleanedModes;
+      const needsUpdate = 'studyNavigatorEnabled' in modes || 'analyticsInterfacePubliclyAccessible' in modes;
+
+      if (needsUpdate) {
+        const cleanedModes = cleanupModes(modes);
+        await setDoc(revisitModesDoc, cleanedModes);
+        return cleanedModes;
+      }
+
+      return modes;
     }
 
     // Else set to default values
--- a/src/storage/engines/SupabaseStorageEngine.ts
+++ b/src/storage/engines/SupabaseStorageEngine.ts
@@ -347,13 +347,20 @@ export class SupabaseStorageEngine extends CloudStorageEngine {
       // get the metadata field from the data object
       const metadata = data[0].data;
       if (metadata) {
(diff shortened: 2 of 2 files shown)


src/storage/engines/FirebaseStorageEngine.ts
```diff
@@ -371,7 +372,10 @@ export class FirebaseStorageEngine extends CloudStorageEngine {
     const revisitModesData = await getDoc(revisitModesDoc);
 
     if (revisitModesData.exists()) {
-      return revisitModesData.data() as Record<REVISIT_MODE, boolean>;
+      const modes = revisitModesData.data() as Record<string, boolean>;
+      const cleanedModes = cleanupModes(modes);
+      await setDoc(revisitModesDoc, cleanedModes);
```
JackWilb: This is guaranteed to send a request every time we pull the modes. We should design this in a way to minimize calls to the DB. That is, we should add a check here and only do the cleanup if necessary

src/storage/engines/LocalStorageEngine.ts
```diff
@@ -170,16 +170,17 @@ export class LocalStorageEngine extends StorageEngine {
     // Get the modes
     const modes = await this.studyDatabase.getItem(key) as Record<REVISIT_MODE, boolean> | null;
     if (modes) {
-      return modes;
+      const cleanedModes = cleanupModes(modes as Record<string, boolean>);
+      await this.studyDatabase.setItem(key, cleanedModes);
+      return cleanedModes;
```
JackWilb: Setting here is low overhead so we can leave it as is

src/storage/engines/SupabaseStorageEngine.ts
```diff
@@ -347,7 +347,13 @@ export class SupabaseStorageEngine extends CloudStorageEngine {
...
       const metadata = data[0].data;
       if (metadata) {
-        return metadata;
+        const cleanedModes = cleanupModes(metadata as Record<string, boolean>);
+        await this.supabase
+          .from('revisit')
+          .update({ data: cleanedModes })
+          .eq('studyId', `${this.collectionPrefix}${studyId}`)
+          .eq('docId', 'metadata');
+        return cleanedModes;
```
JackWilb: This is guaranteed to send a request every time we pull the modes. We should design this in a way to minimize calls to the DB. That is, we should add a check here and only do the cleanup if necessary

JackWilb (changes requested):

PR #1159, opened 2026-03-19
Add independent webcam recording support across capture, storage, and analysis replay

### LINKED ISSUE #1015: Enable webcam recording
From Lingyi Fu:

> We plan to use this platform for think-aloud sessions to evaluate the chatbot. Capturing participants’ facial expressions while they interact with the chatbot is an important part for the evaluation. We expect to conduct the study in approximately one to two months and would greatly appreciate any suggestions you may have.

This feature addition would need us to enable webcam recording, perhaps simultaneously with the screen/audio recording. We'll need to handle the multiple streams, and should probably, for now, keep all these streams separate.

In the analysis UI we could overlay the streams or we could play them side by side. There will likely be sync issues that we need to address.

JackWilb:
Closes #1015

This PR adds webcam recording as a first-class study feature, independent from screen recording and audio recording.

It introduces recordWebcam study/component config support, extends the recording runtime to manage separate screen/webcam/mic streams and per-trial assets, adds storage/download support for webcam recordings across all storage engines, and updates analysis replay to handle screen-only, webcam-only, or synchronized side-by-side screen + webcam playback. It also adds a webcam-only recording library/component, updates the existing screen-recording setup flow for combined capture, and includes example studies plus regenerated JSON schemas.

TODOs:
- [ ] ...

### DIFF
--- a/public/demo-screen-recording/assets/introduction.md
+++ b/public/demo-screen-recording/assets/introduction.md
@@ -1,8 +1,8 @@
-This example demonstrates how to capture screen recordings for components where provenance tracking is difficult to implement. In this case, we illustrate the process using simple website components that are recorded directly from the screen.
+This example demonstrates how to capture screen and webcam recordings for components where provenance tracking is difficult to implement. In this case, we illustrate the process using simple website components that are recorded directly from the screen and webcam.
 
 ## Relevant files:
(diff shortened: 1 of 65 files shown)


JackWilb:
@codex review

PR #1220, opened 2026-04-29
Tutorial fixes

mcnuttandrew:
- return the .env
- return example-brush
- fix the install issue


mcnuttandrew:
something is borked pulling from jack's stuff, closing

PR #1223, opened 2026-05-07
Cache waveform peaks and use direct URLs for faster replay loads

### LINKED ISSUE #1174: replay videos should load faster
Looking at participant replays on video can be super duper slow because we need to download the whole file, often multiple times. One improvement could be to precompute and save the wave form so we don't have to download it to do so.

hilsonshrestha:
Closes #1174 

Audio and screen recording replays were slow because WaveSurfer had to fully decode audio on every load to render the waveform. This PR caches decoded waveform peaks to storage after the first load so subsequent visits skip decoding. It also replaces entire blob fetches of audio and video with direct URL calls so that the media player can buffer stream.

PR also shows a warning when participant used a browser different.

PR adds a Firebase Cloud Function that runs ffmpeg on uploaded screen recordings to re-encode them as webm using a stream copy (no re-encoding when the source is already webm-compatible).

image

x


functions/src/index.ts
```diff
@@ -0,0 +1,92 @@
...
+    if (!SCREEN_RECORDING_PATH.test(filePath)) return;
+
+    // Prevent re-trigger loop: uploading back to the same path fires onObjectFinalized again
+    if (event.data.metadata?.converted === 'true') {
+      logger.info(`Skipping already-converted: ${filePath}`);
+      return;
+    }
+
+    const fileName = path.basename(filePath);
+    const tmpInput = path.join(os.tmpdir(), fileName);
```
JackWilb, line 51: [P1] Second-generation functions can process events concurrently, but these temp paths use only the object basename. Identically named participant/task recordings from different studies can overwrite each other and corrupt the converted upload. Please create a unique temp directory per event (for example with `fs.mkdtemp`) or include a full-path hash/event ID in both paths, then clean it up in `finally`.

src/components/screenRecording/ScreenRecordingReplay.tsx
```diff
@@ -45,7 +45,7 @@ export function ScreenRecordingReplay() {
             if (!participantId) {
               throw new Error('Participant ID is required to load audio');
             }
-            const url = await storageEngine.getScreenRecording(identifier, participantId);
+            const url = await storageEngine.getScreenRecordingUrl(identifier, participantId);
```
JackWilb, line 48: [P1] This method was renamed to `getScreenRecordingUrl`, but the sibling spec still mocks `getScreenRecording`. The focused Vitest run fails and emits unhandled rejections because the new method is missing. Please update the null, success, and video-src test mocks to expose `getScreenRecordingUrl`; CI needs to pass before merge.

JackWilb (changes requested):
Two blocking issues need to be addressed before merge. CI also needs to be fixed: the committed ScreenRecordingReplay unit tests still mock the old storage method, causing the Vitest workflow to fail with one failed assertion and unhandled rejections.

PR #1260, opened 2026-06-10
Visualize study browser skip branching

### LINKED ISSUE #1257: Improve visualization of skip / branching
Right now, the study browser doesn't show anything about skips / branching. 

We should visualization that there's a branching condition at a component, and where it goes to. Probably with the conditional information on mouse over.

JackWilb:
## Summary
- add Participant View-only branch indicators for blocks with skip logic in the Study Browser
- show human-readable skip summaries in one block-level tooltip
- distinguish branch-skipped rows from rows excluded by randomization or condition filtering

## Tests
- yarn unittest --run src/components/interface/StepsPanel.spec.tsx
- yarn unittest --run src/utils/handleSkipLogic.spec.ts src/store/hooks/useNextStep.spec.ts src/store/hooks/useNextStep.spec.tsx
- yarn typecheck
- yarn unittest --run
- yarn lint
- git diff --check

Closes #1257

### Documentation

- [x] Done
- [ ] N/A


src/components/interface/StepsPanel.tsx
jaykim1213, line 1: Small naming suggestion: could we remove the `branch` term and use `skip`/`skipped` instead?

- `getBranchSkippedTrialOrders` -> `getSkippedTrialOrders`
- `branchSkippedTrialOrders` -> `skippedTrialOrders`
- `isBranchSkipped` -> `isSkipped`
- `isBranchSkippedBlock` -> `isSkippedBlock`
- `exclusionReason?: 'excluded' | 'branch'` -> `exclusionReason?: 'excluded' | 'skipped'`

src/components/interface/StepsPanel.tsx
```diff
@@ -786,6 +819,16 @@ export function StepsPanel({
...
                         </Tooltip>
                       )}
+                      {!isComponent && skipSummaries && skipSummaries.length > 0 && (
+                        <Tooltip
+                          label={skipSummaries.join(' ')}
+                          multiline
+                          position="right"
+                          withArrow
+                        >
+                          <IconArrowsSplit size={16} style={{ marginRight: 4, flexShrink: 0 }} color="teal" />
```
jaykim1213: - @alexsb suggested changing the icon here. I was thinking about `<IconArrowForward>` which looks like this 

Image

- Small thing: `teal` looks almost the same as the green used for conditional blocks. Could we use a more distinct color like light blue/light green, or keep green but different `opacity` to make the visual difference?

src/components/interface/StepsPanel.tsx
```diff
@@ -811,6 +854,11 @@ export function StepsPanel({
                       >
                         {label}
                       </Text>
+                      {isComponent && isBranchSkipped && (
+                        <Badge ml={5} color="teal" variant="light">
```
jaykim1213:
```suggestion
                        <Badge ml={5} color="gray" variant="light">
```

Can we use gray like we do for excluded components?

src/components/interface/StepsPanel.tsx
```diff
@@ -829,8 +877,8 @@ export function StepsPanel({
                         </Badge>
                       )}
                       {!isComponent && isExcluded && (
-                        <Badge ml={5} color="gray" variant="light">
-                          Excluded
+                        <Badge ml={5} color={isBranchSkipped ? 'teal' : 'gray'} variant="light">
```
jaykim1213:
```suggestion
                        <Badge ml={5} color="gray" variant="light">
```

I think using gray here is okay since a skipped component is also kind of like an excluded component.

src/components/interface/StepsPanel.tsx
jaykim1213, line 1: Things to discuss:

Right now, we show the `Skipped` badge on skipped components and blocks, but we only show the `Excluded` badge on excluded blocks, not excluded components. So `Skipped` and `Excluded` behave a little differently.

I think we have three options:

1. Keep it as-is

Image

Image

 - probably the easiest, but is not consistent

2. Show `Excluded` badges on components too, so it matches `Skipped`

Image

- This is more consistent, but it can create a lot of badges

3. Only show component-level badges when the parent block does not already show the same badge

- If the whole block is skipped/excluded, show the badge on the block only
- If just one component inside an active block is skipped/excluded, show the badge on that component
- Needs a bit of more implementation to check whether the parent block already has the badge

Image

Image

src/components/interface/StepsPanel.utils.ts
```diff
@@ -14,3 +21,204 @@ export function getDynamicComponentsForBlock(
...
+  if (condition.check === 'block') {
+    const result = condition.condition === 'numCorrect' ? 'correct' : 'incorrect';
+    return `If this block has ${pluralize(condition.value, result, result)} component ${condition.value === 1 ? 'response' : 'responses'}, skip to ${target}.`;
+  }
+
+  const result = condition.condition === 'numCorrect' ? 'correct' : 'incorrect';
+  return `If ${formatReference(condition.name)} has ${pluralize(condition.value, result, result)} repeated ${condition.value === 1 ? 'response' : 'responses'}, skip to ${target}.`;
+}
+
+function conditionComponentIsPossible(condition: SkipConditions[number], sequence: Sequence) {
```
jaykim1213: I feel like `conditionComponentIsPossible` could be confused with the URL condition feature we have. Could we include `skip` in the name  like `shouldShowSkipConditionForBlock` or something else?

src/components/interface/StepsPanel.utils.ts
```diff
@@ -14,3 +21,204 @@ export function getDynamicComponentsForBlock(
...
+      };
+    }
+    return acc;
+  }, {});
+}
+
+function getConditionTargetIndex(condition: SkipConditions[number], sequence: Sequence, flatSequence: string[]) {
+  if (condition.to === 'end') {
+    return flatSequence.length;
+  }
```
jaykim1213: I think this might point to the index after the last item

```suggestion
if (condition.to === 'end') {
  const endIndex = flatSequence.indexOf('end');
  return endIndex === -1 ? flatSequence.length : endIndex;
}
```

or we could do 


```suggestion
  if (condition.to === 'end') {
    return flatSequence.length - 1;
  }
```

src/components/interface/StepsPanel.utils.ts
jaykim1213, line 1: `StepsPanel.utils.ts` seems to have some skip logic that is similar to `useNextStep`. I think we could extract the clearly shared pieces into something like `skipLogic.ts` or `skipConditions.ts`.

Here is the comment from Codex:

- `getBranchSkippedTrialOrders` does a similar skip-evaluation process as `useNextStep`: it finds the current block, collects skip conditions, checks whether any condition is triggered, and resolves the skip target. This part may still need to stay in `StepsPanel.utils.ts` because it reconstructs skipped rows from saved participant answers.
- `conditionIsTriggered` is similar to the condition-checking logic inside `useNextStep`: it filters the answers for the current block and decides whether a `response`, `responses`, `block`, or `repeatedComponent` condition passed.
- `getAnswersForSkipEvaluation` is similar to `answersForSkipEvaluation` in `useNextStep`: it builds the answer object used to evaluate skip conditions. One difference is that `StepsPanel` builds from saved participant answers, while `useNextStep` also includes the current unsaved answer.
- `getConditionTargetIndex` is similar to the target-resolution logic in `useNextStep`: it converts `condition.to` into either a component index, block index, or end target.

jaykim1213 (changes requested):

src/components/interface/StepsPanel.tsx
alexsb: Option 3 is the nicest; Option 2 is fine too – I assume rendering wouldn't be an issue for large studies?

JackWilb: Addressed in 2176b815. Removed the branch terminology in this path: `getBranchSkippedTrialOrders` is now `getSkippedTrialOrders`, local variables use skipped naming, and `exclusionReason` now uses `skipped` instead of `branch`.

Verified with `yarn unittest --run src/components/interface/StepsPanel.spec.tsx src/components/interface/tests/StepsPanel.spec.tsx src/store/hooks/tests/useNextStep.spec.ts`, `yarn typecheck`, `yarn lint`, and `yarn build`.

src/components/interface/StepsPanel.tsx
JackWilb: Addressed in 2176b815. Skipped and excluded status badges now both use gray light badges.

Verified with the focused unit tests, `yarn typecheck`, `yarn lint`, and `yarn build`.

src/components/interface/StepsPanel.tsx
JackWilb: Addressed in 2176b815. Swapped the skip-condition icon to `IconArrowForward` and changed its color from teal to blue so it is more distinct from the green conditional-block icon.

Verified with the focused unit tests, `yarn typecheck`, `yarn lint`, and `yarn build`.

src/components/interface/StepsPanel.tsx
JackWilb: Addressed in 2176b815. The component-level skipped badge now uses the same gray light badge treatment as excluded items.

Verified with the focused unit tests, `yarn typecheck`, `yarn lint`, and `yarn build`.

src/components/interface/StepsPanel.tsx
JackWilb: Addressed in 2176b815 using Option 3. The row metadata now tracks the parent exclusion reason, so component-level `Skipped` or `Excluded` badges only render when the parent block does not already show the same badge. I also added UI tests covering direct excluded rows versus children under an excluded block.

Verified with the focused unit tests, `yarn typecheck`, `yarn lint`, and `yarn build`.

src/components/interface/StepsPanel.utils.ts
JackWilb: Addressed in 2176b815. Extracted the shared skip-condition pieces into `src/utils/skipConditions.ts`: condition evaluation, target resolution, correctness helpers, and stored-answer preparation. `useNextStep` and `StepsPanel.utils.ts` now both use that shared logic, while `StepsPanel.utils.ts` keeps the persisted-answer reconstruction that is specific to the study browser.

Verified with the focused unit tests, `yarn typecheck`, `yarn lint`, and `yarn build`.

src/components/interface/StepsPanel.utils.ts
JackWilb: Addressed in 2176b815. Renamed `conditionComponentIsPossible` to `shouldShowSkipConditionForBlock` so this helper clearly belongs to skip-condition display logic.

Verified with the focused unit tests, `yarn typecheck`, `yarn lint`, and `yarn build`.

src/components/interface/StepsPanel.utils.ts
JackWilb: Addressed in 2176b815. `condition.to === "end"` now resolves to the explicit `end` sentinel index when present, with `flatSequence.length` kept only as a defensive fallback if a sequence lacks the sentinel. I also added a regression test so the `end` row is not marked skipped.

Verified with the focused unit tests, `yarn typecheck`, `yarn lint`, and `yarn build`.

jaykim1213 (approved):
Thank you Jack! 

One small thing: the skip icon is blue, which overlaps with the library import icon. I don’t think it’s a big deal, but if we want them to be different, we could use `purple` instead. Otherwise, blue looks okay too! 😄

PR #1261, opened 2026-06-10
Fix replay task ordering

### LINKED ISSUE #1240: Fix task ordering in replay when participants use previous buttons
In Participant View replay timelines, task ordering can become incorrect when a participant navigates backward using the previous button.

The `startTime` / `endTime` values can affect the replay timeline ordering and make tasks appear out of sequence. They should always show up in the their sequence order

---

## Specification

### Summary
Participant View replay timelines should render tasks in study sequence order even when a participant used Previous and the saved `startTime` / `endTime` values are no longer monotonic. This should be a replay rendering fix, not a stored-data migration.

### Decisions
- Sort replay task entries by parsed `trialOrder`, using sequence step first, dynamic `funcIndex` second, and task identifier as a stable tie-breaker.
- Keep `startTime` / `endTime` as the source for timeline scale, task bar position, bar width, duration, and browsed-away overlays.
- Include incomplete answers with `startTime === 0` in the same sequence-order sort rather than appending them after completed tasks.
- Sort dynamic block trials such as `3_0`, `3_1`, and `3_2` by parent step and then `funcIndex`; if both a non-dynamic `3` and dynamic children exist, the non-dynamic step sorts before its dynamic children.
- Do not modify saved participant answer data.
- Do not change analytics summaries or exported data ordering unless they share the same visible Participant View replay timeline bug.

### Acceptance Criteria
- [ ] Participant View replay timeline renders task labels/tasks in study sequence order when saved answers have non-monotonic `startTime` values caused by Previous navigation.
- [ ] Dynamic/repeated task answers sort by `trialOrder` step and `funcIndex`.
- [ ] Incomplete answers do not get appended out of sequence solely because `startTime === 0`.
- [ ] Timestamp-based bar scale, bar widths, and browsed-away overlays continue to use recorded `startTime` / `endTime`.
- [ ] Clicking a task in the replay timeline still opens the correct trial using that task's `trialOrder`.
- [ ] Focused unit coverage is added for the ordering helper or `AllTasksTimeline` behavior using out-of-order timestamps.

### Out of Scope
- Rewriting participant answer persistence.
- Changing how `startTime` / `endTime` are collected.
- Changing exported analytics semantics.
- Adding new libraries.

### Implementation Notes
- The likely control point is `src/analysis/individualStudy/replay/AllTasksTimeline.tsx`, which currently sorts completed tasks by `startTime` while incomplete tasks use `trialOrder`.
- Prefer reusing the existing `parseTrialOrder` utility instead of adding ad-hoc parsing logic.
- Existing think-aloud replay code already orders answers by parsed `trialOrder` with an identifier tie-breaker; use that as the behavioral reference.

### Testing Notes
- Add a focused Vitest test for out-of-order timestamps where `trialOrder` is still ordered, including at least one dynamic `trialOrder` case.
- Confirm timestamp-driven visual metrics still use `startTime` / `endTime` after the ordering change.

JackWilb:
## Summary
- sort Participant View replay timeline tasks by parsed `trialOrder` instead of recorded timestamps
- preserve timestamp-based scale, bar positioning, widths, and browsed-away overlays
- add focused Vitest coverage for out-of-order timestamps, dynamic trials, incomplete answers, and tie-breaking

Fixes #1240.

## Testing
- `yarn unittest src/analysis/individualStudy/replay/taskOrdering.spec.ts --run`
- `yarn typecheck`
- `yarn lint`

### Documentation
Tracking issue: https://github.com/revisit-studies/reVISit-studies.github.io/issues/238

- [x] Done
- [ ] N/A

### DIFF
--- a/src/analysis/individualStudy/replay/AllTasksTimeline.tsx
+++ b/src/analysis/individualStudy/replay/AllTasksTimeline.tsx
@@ -8,22 +8,20 @@ import {
 import { ParticipantData } from '../../../storage/types';
 import { SingleTaskLabelLines } from './SingleTaskLabelLines';
 import { SingleTask } from './SingleTask';
-import { StoredAnswer, StudyConfig } from '../../../parser/types';
+import { StudyConfig } from '../../../parser/types';
 import { getComponentAnswerStatus } from '../../../utils/correctAnswer';
 import { parseConditionParam } from '../../../utils/handleConditionLogic';
 import { studyComponentToIndividualComponent } from '../../../utils/handleComponentInheritance';
+import {
+  compareReplayAnswerEntries,
+  orderedReplayAnswerEntries,
+} from './taskOrdering';
 
 const LABEL_GAP = 25;
 const CHARACTER_SIZE = 8;
 
 const margin = {
   left: 20, top: 20, right: 20, bottom: 20,
-};
-
-const sortedTaskNames = (a: [string, StoredAnswer], b: [string, StoredAnswer]) => {
-  const splitA = a[1].trialOrder.split('_');
-  const splitB = b[1].trialOrder.split('_');
-  return splitA[0] === splitB[0] ? +splitA[1] - +splitB[1] : +splitA[0] - +splitB[0];
 };
 
 export function AllTasksTimeline({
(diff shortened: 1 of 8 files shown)


JackWilb:
Enable a toggle for chronological by answer time

PR #1262, opened 2026-06-10
Add uniform participant timeline mode

### LINKED ISSUE #1239: Uniform Width for tasks in participant view
It would be good to have the option to switch the participant view over to "uniform width" for each component, so that we don't have weird outliers and can instead focus on looking at results by participants. This should be a toggle that we can click on this participant view visualization.

Here's the current UI
Image

Instead, each "bar" would be of equal width; space-filling, but also with a minimum width (something reasonable, like 40-60px).

Paired with that, we could color bars by correctness in both modes.

Ccing @ZachCutler04 since we talked about this before.

---

## Specification

### Summary
Add a Participant View toggle that switches expanded participant-row task timelines between the current time-proportional layout and a uniform-width layout. The target visualization is the detail-panel timeline rendered by `src/analysis/individualStudy/replay/AllTasksTimeline.tsx`, not the Trial Stats charts.

### Decisions
- Scope the feature to the `Participant View` table detail-panel timeline.
- Add a compact Mantine switch or segmented control in the Participant View table toolbar, near the existing row actions.
- Apply the selected timeline mode globally within Participant View for the current page session, so expanded participant rows use the same mode.
- Do not persist the selected mode to storage or user settings in the first implementation.
- In uniform-width mode, keep the existing task ordering: completed tasks in their current timeline/order sequence, followed by incomplete tasks sorted by trial order.
- In uniform-width mode, each task receives equal visual width. Use a minimum width of about `48px`; when `taskCount * 48px` exceeds the available row width, allow the timeline to scroll horizontally instead of shrinking bars below the minimum.
- Preserve existing labels, label connector lines, hover expansion, tooltips, recording icons, and click-to-trial navigation in both modes.
- Color task bars by correctness in both time-proportional and uniform-width modes:
  - Correct tasks: green.
  - Incorrect tasks: red.
  - Incomplete tasks: neutral light gray or orange-tinted neutral.
  - Tasks without a correct-answer definition: neutral gray.
- Hover styling may add emphasis or an outline, but should not obscure the correctness meaning.
- Keep duration-based browsed-away markers in time-proportional mode. Hide or disable those markers in uniform-width mode so they do not imply duration when width is no longer time-based.

### Acceptance Criteria
- [ ] Participant View has a visible control to switch expanded task timelines between time-proportional and uniform-width modes.
- [ ] Time-proportional mode preserves the current duration-based bar width behavior.
- [ ] Uniform-width mode renders each task bar at equal width, fills available horizontal space when possible, and uses an approximately `48px` minimum width.
- [ ] Uniform-width timelines remain readable when many tasks are present by using horizontal overflow rather than shrinking bars below the minimum width.
- [ ] Correct-answer tasks are colored by correctness in both modes.
- [ ] Incomplete tasks and tasks without correct-answer definitions remain visually distinguishable from correct and incorrect tasks.
- [ ] Existing task hover behavior, tooltips, labels, recording icons, and click navigation still work in both modes.
- [ ] Duration-specific browsed-away markers remain available in time-proportional mode and do not render misleadingly in uniform-width mode.
- [ ] The feature does not change exported participant data, study config parsing, Trial Stats visualizations, or persisted analysis data.

### Out of Scope
- Persisting the selected timeline mode across page reloads or users.
- Changing Trial Stats charts.
- Changing exported participant data or storage schemas.
- Adding new libraries.
- Reworking the full Participant View table layout.

### Implementation Notes
- Likely implementation targets are `src/analysis/individualStudy/table/TableView.tsx`, `src/analysis/individualStudy/replay/AllTasksTimeline.tsx`, `src/analysis/individualStudy/replay/SingleTask.tsx`, and `src/analysis/individualStudy/replay/SingleTaskLabelLines.tsx`.
- Consider factoring timeline position and width calculation into a small helper so time-proportional and uniform-width behavior can be tested without relying only on visual tests.
- Reuse the existing correctness calculation in `AllTasksTimeline`, which already derives `isCorrect` and `hasCorrect` before rendering each `SingleTask`.

### Testing Notes
- Add focused unit coverage for any extracted timeline layout helper, including the minimum-width overflow case.
- Add or extend Playwright coverage for Participant View to verify the mode toggle is visible and changes expanded timeline bar widths while preserving task navigation/hover affordances where practical.

JackWilb:
image

Screenshot 2026-06-22 at 3 04 01 PM

## Summary
- add a Participant View timeline mode control for Time vs Uniform layouts
- render uniform task bars with a 48px minimum width and horizontal overflow when needed
- color timeline task bars by correctness/incomplete/no-correct-answer state and hide duration-only browsed-away markers in uniform mode
- add focused tests for uniform timeline sizing and ordering

## Verification
- `yarn unittest src/analysis/individualStudy/replay/timelineLayout.spec.ts src/analysis/individualStudy/replay/taskOrdering.spec.ts --run`
- `yarn typecheck`
- `yarn lint`
- `yarn build`
- Browser smoke loaded `http://localhost:8080/analysis/stats/demo-choropleth-map/table` with no console errors; local data had no participant rows, so the table toolbar was not mounted for visual interaction.

Fixes #1239

### Documentation
Tracking issue: https://github.com/revisit-studies/reVISit-studies.github.io/issues/238

- [x] Done
- [ ] N/A

### DIFF
--- a/src/analysis/individualStudy/replay/AllTasksTimeline.tsx
+++ b/src/analysis/individualStudy/replay/AllTasksTimeline.tsx
@@ -226,25 +226,24 @@ export function AllTasksTimeline({
     });
   }, [xScale, maxHeight, participantData.answers, timelineMode]);
 
-  const hoveredTask = tasks.find((task) => task.identifier === hoveredTaskIdentifier);
-  const nonHoveredTasks = tasks.filter((task) => task.identifier !== hoveredTaskIdentifier);
-
   return (
-    <Center>
-      <Stack gap={15} style={{ width: '100%' }}>
-        <Box style={{ width: '100%', overflowX: timelineMode === 'uniform' ? 'auto' : 'visible', overflowY: 'visible' }}>
+    <Center style={{ width: '100%', minWidth: 0 }}>
+      <Stack gap={15} style={{ width: '100%', minWidth: 0 }}>
+        <Box style={{
+          width: '100%', maxWidth: '100%', minWidth: 0, overflowX: timelineMode === 'uniform' ? 'auto' : 'visible', overflowY: 'visible',
+        }}
+        >
           <svg
             onMouseLeave={() => setHoveredTaskIdentifier(null)}
             style={{
               width: timelineWidth,
-              minWidth: timelineWidth,
               height: maxHeight,
+              display: 'block',
               overflow: 'visible',
             }}
           >
             {tasks.map((t) => t.line)}
-            {nonHoveredTasks.map((t) => t.label)}
-            {hoveredTask?.label}
+            {tasks.map((t) => t.label)}
             {browsedAway}
           </svg>
         </Box>
--- a/src/analysis/individualStudy/replay/AllTasksTimeline.tsx
+++ b/src/analysis/individualStudy/replay/AllTasksTimeline.tsx
@@ -3,7 +3,7 @@ import {
 } from 'react';
 import * as d3 from 'd3';
 import {
-  Center, Stack, Tooltip, Text,
+  Box, Center, Stack, Tooltip, Text,
 } from '@mantine/core';
 import { ParticipantData } from '../../../storage/types';
 import { SingleTaskLabelLines } from './SingleTaskLabelLines';
@@ -16,6 +16,11 @@ import {
   compareReplayAnswerEntries,
   orderedReplayAnswerEntries,
 } from './taskOrdering';
+import {
+  getUniformTimelineMetrics,
+  orderedUniformReplayAnswerEntries,
+  TimelineMode,
+} from './timelineLayout';
 
 const LABEL_GAP = 25;
 const CHARACTER_SIZE = 8;
@@ -25,8 +30,8 @@ const margin = {
 };
 
 export function AllTasksTimeline({
-  participantData, width, studyId, studyConfig, maxLength,
-}: { participantData: ParticipantData, width: number, studyId: string, studyConfig: StudyConfig | undefined, maxLength: number | undefined }) {
+  participantData, width, studyId, studyConfig, maxLength, timelineMode = 'time',
+}: { participantData: ParticipantData, width: number, studyId: string, studyConfig: StudyConfig | undefined, maxLength: number | undefined, timelineMode?: TimelineMode }) {
   const [hoveredTaskIdentifier, setHoveredTaskIdentifier] = useState<string | null>(null);
 
   const percentComplete = useMemo(() => {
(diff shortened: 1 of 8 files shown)


src/analysis/individualStudy/table/TableView.tsx
jaykim1213, line 1: Image

Can we move the Time/Uniform control before the Reject/Undo Reject Participant button so the control doesn’t shift when participants are selected? 
Also the sizes of control and buttons are different, so we could either change the Reject/Undo Reject button to `xs` or make the Time/Uniform control `sm` so their sizes match

src/analysis/individualStudy/replay/SingleTask.tsx
```diff
@@ -16,6 +16,26 @@ const LABEL_MAX_WIDTH = 160;
...
+  incomplete,
+  hasCorrect,
+  isCorrect,
+}: {
+  incomplete: boolean,
+  hasCorrect: boolean,
+  isCorrect: boolean,
+}) {
+  if (incomplete) {
+    return '#ffe8cc';
```
jaykim1213:
```suggestion
    return '#e9ecef';
```

Looks great. 
I just added small color suggestion: with the current changes, the in-progress is marked as yellow. I’d keep light gray/light gray for incomplete/complete tasks like we already do. With yellow, it feels like the task has some meaningful value

Current 

Image

Suggestion

Image

src/analysis/individualStudy/replay/SingleTask.tsx
```diff
@@ -16,6 +16,26 @@ const LABEL_MAX_WIDTH = 160;
...
+  }
+
+  if (!hasCorrect) {
+    return '#dee2e6';
```
jaykim1213:
```suggestion
    return 'lightgray';
```

src/analysis/individualStudy/replay/SingleTask.tsx
```diff
@@ -16,6 +16,26 @@ const LABEL_MAX_WIDTH = 160;
...
+  }
+
+  return isCorrect ? '#40c057' : '#fa5252';
```
jaykim1213: For correct and incorrect tasks, I'd suggest making the bar colors a little lighter to make it softer. We can also ask for feedback on Slack

Current
Image

1.  `return isCorrect ? '#cfe3d1' : '#efd2d2';`

Image

2. `return isCorrect ? '#bfe8c6' : '#f3c1c1';`

Image

3. `return isCorrect ? '#b3d9ba' : '#e8b6b6';`

Image

src/analysis/individualStudy/replay/AllTasksTimeline.tsx
jaykim1213, line 1: The horizontal timeline should stay inside the table. Each timeline should have its own horizontal scroll instead of making the whole table scroll horizontally. 

Image

jaykim1213, line 1: There is a bug with the hovering: the stroke sometimes stays there after moving the cursor away from the task. It makes it look like two tasks are selected

Image

src/analysis/individualStudy/replay/AllTasksTimeline.tsx
```diff
@@ -160,10 +188,14 @@ export function AllTasksTimeline({
     });
 
     return allElements;
-  }, [participantData.answers, participantData.participantId, incompleteXScale, xScale, studyConfig, maxHeight, studyId, conditionParam, hoveredTaskIdentifier]);
+  }, [participantData.answers, participantData.participantId, incompleteXScale, xScale, studyConfig, maxHeight, studyId, conditionParam, hoveredTaskIdentifier, timelineMode]);
 
   // Find entries of someone browsing away. Show them
   const browsedAway = useMemo(() => {
+    if (timelineMode === 'uniform') {
```
jaykim1213, line 198: I believe browsed away is hidden in uniform mode because browsed away is timestamp based. I think we should mention this in the documentation if this gets merged in

src/analysis/individualStudy/replay/AllTasksTimeline.tsx
```diff
@@ -192,20 +224,30 @@ export function AllTasksTimeline({
...
-          {nonHoveredTasks.map((t) => t.label)}
-          {hoveredTask?.label}
-          {browsedAway}
-        </svg>
+        <Box style={{ width: '100%', overflowX: timelineMode === 'uniform' ? 'auto' : 'visible', overflowY: 'visible' }}>
+          <svg
+            onMouseLeave={() => setHoveredTaskIdentifier(null)}
+            style={{
+              width: timelineWidth,
+              minWidth: timelineWidth,
```
jaykim1213: Do we need `minWidth` here if `width` is already `timelineWidth`?

jaykim1213 (changes requested):

src/analysis/individualStudy/table/TableView.tsx
JackWilb: Moved the Time/Uniform segmented control before the reject actions and changed it to size="sm" so selecting participants no longer shifts the control and the sizing matches the buttons.

src/analysis/individualStudy/replay/SingleTask.tsx
JackWilb: Changed incomplete task fill to #e9ecef so in-progress tasks read as neutral light gray instead of yellow.

src/analysis/individualStudy/replay/AllTasksTimeline.tsx
JackWilb: Fixed the stale hover highlight by keeping task SVG render order stable and using pointer leave/cancel handling, so the stroke clears when the cursor leaves the task.

src/analysis/individualStudy/replay/AllTasksTimeline.tsx
JackWilb: Removed the redundant svg minWidth; width: timelineWidth remains, and the scroll wrapper now owns overflow behavior.

src/analysis/individualStudy/replay/SingleTask.tsx
JackWilb: Softened the correct/incorrect task fills to #bfe8c6 and #f3c1c1, matching option 2 from the suggestions.

src/analysis/individualStudy/replay/AllTasksTimeline.tsx
```diff
@@ -192,22 +227,33 @@ export function AllTasksTimeline({
...
+    >
+      <svg
+        onMouseLeave={() => setHoveredTaskIdentifier(null)}
+        style={{
+          width: timelineWidth,
+          height: maxHeight,
+          display: 'block',
+          overflow: 'visible',
+        }}
+      >
```
jaykim1213, line 255: Image

Hovered labels are not rendered on top, so the expanded label can still get overlapped. We could reuse the `hoveredTask` we had before. 

One thing is that I tried rendering with `hoveredTask` locally, but it brought back the old bug where the hover state does not clear correctly and makes it look like two tasks are selected. To solve this I think we could also add `onPointerLeave` and `onPointerCancel` to clear `hoveredTaskIdentifier` when the cursor leaves the timeline

jaykim1213 (changes requested):

PR #1274, opened 2026-06-12
Re-save missing active config and handle missing configs in Tidy export

### LINKED ISSUE #1273: Tidy download not working when stored config is missing
Tidy download not working when stored config is missing

Image

---

## Summary
Deleting live data from Data Management is expected to delete stored configs along with participants, media, provenance, and realtime data. The bug is that the app can later trust an existing config hash and skip re-saving the active config even when the actual stored config object is missing. Tidy CSV export should also handle older or broken participant data whose `participantConfigHash` cannot be resolved.

## Decisions
- Treat deletion of stored configs during “Delete Live Data” as expected current behavior.
- Fix `saveConfig` so it verifies the active config object actually exists before skipping the write. If the hash matches but the stored config object is missing, re-save the active config.
- Keep Tidy CSV defensive: when `getAllConfigsFromHash` succeeds but one or more participant config hashes cannot be resolved, continue building rows from the participant data that is still available.
- Do not automatically fall back to the current active study config for participants with missing historical configs, because task metadata, prompts, correct answers, and response definitions may differ from what the participant saw.
- Represent participants with missing configs using participant-derived tidy rows where possible. Config-derived columns may be blank.
- If every selected participant has a missing config, still produce a CSV containing headers and participant-derived rows.
- Continue transcript fetching for participants with missing configs when transcript export is selected, because transcripts are participant-derived and do not require the stored config.
- If `getAllConfigsFromHash` rejects entirely, keep treating that as an export-blocking storage error and show the existing modal error state.

## Acceptance Criteria
- [ ] `saveConfig` re-saves the active config when the stored current config hash matches but `configs/{hash}` is missing.
- [ ] `saveConfig` still avoids unnecessary writes when both the current config hash and stored config object are present.
- [ ] Opening the Tidy CSV Exporter no longer crashes or blocks indefinitely when at least one selected participant references a missing stored config.
- [ ] Participants with available configs continue to produce normal tidy task rows.
- [ ] Participants with missing configs still produce participant-derived tidy rows where possible, including fields such as `participantId`, `trialId`, `responseId`, `answer`, timing fields, parameters, `configHash`, participant tags, metadata, and window events.
- [ ] Config-derived fields such as `description`, `instruction`, `responsePrompt`, `meta`, `responseMin`, and `responseMax` may be blank for participants whose stored config cannot be loaded.
- [ ] The Tidy CSV Exporter modal shows a non-blocking warning with the number of affected participants when missing configs are detected.
- [ ] No participant task rows are generated from the active study config as a fallback for a missing historical config.
- [ ] JSON participant download behavior remains unchanged.
- [ ] Add Vitest coverage for the missing-config Tidy CSV cases and the `saveConfig` missing-object recovery case.

## Out of Scope
- Changing what “Delete Live Data” deletes.
- Recovering, reconstructing, or backfilling deleted historical study configs.
- Changing JSON participant export behavior.
- Adding new external libraries.

## Implementation Notes
- Delete Live Data calls `removeSnapshotOrLive(studyId, studyId)`, which deletes the live data namespace, including configs, participants, audio, screenRecording, provenance, and realtime data.
- The storage recovery fix belongs in `src/storage/engines/types.ts` inside `saveConfig`: do not return early on a matching hash unless `configs/{hash}` is actually present.
- The Tidy failure path is in `src/components/downloader/DownloadTidy.tsx`, where `getTableData` fetches configs by `participantConfigHash` and previously passed `allConfigs[participant.participantConfigHash]` directly into row generation.
- Build a missing-config count from selected participants after config lookup. Use it for the modal warning.
- For participants with missing configs, pass `undefined` into row generation instead of the missing/empty stored config object.
- Keep total storage failures distinct from partial missing-config results: total rejection should continue to surface as the existing error alert.

## Testing Notes
- Add storage-level coverage where the current config hash exists but the stored config object is deleted, then verify `saveConfig` restores it.
- Add focused Vitest coverage for Tidy table generation where one participant config resolves and one does not.
- Add coverage where all selected participant configs are missing and the resulting CSV preview/download data still contains participant-derived rows.
- Add coverage for Firebase/Supabase-style missing configs that resolve as an empty object.
- Add coverage for stored config objects that exist but are missing `components`.
- Verify that `getTranscription` is still called for participants whose config is missing when transcript export is selected and transcripts are available.

jaykim1213:
Closes #1273

- Re-save the active config if the stored config file is missing
- Make Tidy CSV handle missing configs gracefully (which should not happen anymore but this is a defensive fallback)

TODOs:
- [ ] ...

### Documentation

- [ ] Done
- [x] N/A

### DIFF
--- a/src/components/downloader/DownloadTidy.tsx
+++ b/src/components/downloader/DownloadTidy.tsx
@@ -251,6 +251,10 @@ function participantDataToRows(
     }).flat()], Array.from(newHeaders)];
 }
 
+function hasStoredStudyConfig(config: StudyConfig | null | undefined): config is StudyConfig {
+  return config ? Object.keys(config).length > 0 : false;
+}
+
 export async function getTableData(
   selectedProperties: Property[],
   data: ParticipantDataWithStatus[],
@@ -266,7 +270,7 @@ export async function getTableData(
 
   const allConfigHashes = [...new Set(data.map((part) => part.participantConfigHash))];
   const allConfigs = await storageEngine.getAllConfigsFromHash(allConfigHashes, studyId);
-  const participantsMissingConfig = data.filter((participant) => !allConfigs[participant.participantConfigHash]);
+  const participantsMissingConfig = data.filter((participant) => !hasStoredStudyConfig(allConfigs[participant.participantConfigHash]));
 
   const transcripts: Record<string, string | null> = {};
   const transcriptAvailable = storageEngine.getEngine() === 'firebase' && !!hasAudio;
@@ -298,7 +302,13 @@ export async function getTableData(
     .filter((p) => p !== 'condition' || hasCondition)
     .filter((p) => p !== 'metaData');
   const allData = await Promise.all(data.map(async (participant) => {
-    const partDataToRows = await participantDataToRows(participant, combinedProperties, allConfigs[participant.participantConfigHash], transcripts);
+    const participantConfig = allConfigs[participant.participantConfigHash];
+    const partDataToRows = await participantDataToRows(
+      participant,
+      combinedProperties,
+      hasStoredStudyConfig(participantConfig) ? participantConfig : undefined,
+      transcripts,
+    );
 
     return partDataToRows;
   }));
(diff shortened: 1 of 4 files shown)


JackWilb (approved):
Thanks Jay! Based on what we currently save this is the best we can do, but should we consider saving instruction and prompt from each question into participant answers so we can recover more gracefully?

PR #1289, opened 2026-06-26
Add verification pass for unresolved React components

WilsonGramer:

I couldn't find a related issue, but this is a feature request from Lane (@codementum).

This PR adds a pass to `verifyStudyConfig` that checks whether imported React components exist in `src/public/`. It uses the same logic to fetch modules as ReactComponentController.tsx. If the module doesn't exist, it suggests to check that the component is in `src/public/` and not `public/`.

In demo-react-trrack/config.json, changing the component name to `Missing.tsx` produces this error:

verify-unresolved-react-components

TODOs:
- [ ] Other component types?

### DIFF
--- a/src/parser/parser.ts
+++ b/src/parser/parser.ts
@@ -19,6 +19,7 @@ import {
   shouldWarnForDefaultFirebaseConfig,
   shouldWarnForDefaultSupabaseConfig,
 } from '../utils/defaultStorageConfig';
+import { studyComponentToIndividualComponent } from '../utils/handleComponentInheritance';
 
 const modules = import.meta.glob(
   [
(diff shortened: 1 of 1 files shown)


src/parser/parser.ts
```diff
@@ -311,6 +319,29 @@ function verifyStudyConfig(studyConfig: StudyConfig, importedLibrariesData: Reco
     });
   });
 
+  // Verify that paths to React components exist under the correct base directory
+  if (studyConfig.baseComponents != null) {
```
JackWilb: `baseComponents` is optional, so this guard skips validation of the required `components` object whenever a study does not define any base components. A direct React component with a missing path therefore receives no error. Please remove the outer guard and always iterate both collections using `studyConfig[key] ?? {}`.

src/parser/parser.ts
```diff
@@ -311,6 +319,29 @@ function verifyStudyConfig(studyConfig: StudyConfig, importedLibrariesData: Reco
...
+    for (const key of ['baseComponents', 'components'] as const) {
+      for (const [name, component] of Object.entries(studyConfig[key] ?? {})) {
+        if (
+          'path' in component
+            && component.path != null
+            && component.type === 'react-component'
```
JackWilb: This checks the raw component rather than its resolved inherited shape. A child can inherit `type: react-component` from its base component while overriding `path`; because `component.type` is then undefined here, an unresolved override passes validation. Please merge the referenced base component with the child before checking the effective type and path, as the existing sidebar validation does.

JackWilb (changes requested):
Hi @WilsonGramer. Thanks for the first PR! We really appreciate your help on this effort.

I'm Requesting changes for two validation gaps that allow unresolved React component paths to pass without an error. Please push up the fixes to these and tag me for a re-review

JackWilb (dismissed):
Thanks for the update! This fixes the common direct-component and overridden-path cases. I found one remaining inherited-path edge case that needs a small follow-up.

JackWilb (approved):
Thanks for the follow-up! The original validation issues are addressed, and the remaining theoretical inheritance case is unlikely in practice. This looks good to merge.

PR #1290, opened 2026-07-01
Fix vega replay against missing signals

### LINKED ISSUE #1268: Fix tooltip signal error in demo-vega
I haven’t been able to reproduce it but in `demo-vega`, the replay sometimes shows this error:

Uncaught Error: Unrecognized signal name: "tooltip" at error (chunk-V7TQW4PD.js?v=254fca26:3109:9) at lookupSignal (chunk-V7TQW4PD.js?v=254fca26:24209:61) at View.signal (chunk-V7TQW4PD.js?v=254fca26:24273:16) at VegaController.tsx:157:13

---

## Specification

### Summary

Fix the intermittent `demo-vega` replay crash by making Vega replay tolerate provenance events whose signal name does not exist in the currently mounted Vega view. The demo appears to expose a replay/controller contract bug rather than an invalid study config: `vegademo1` defines a real `tooltip` signal, while `vegademo2` uses `tooltip` only as a data field and defines different replayable signals.

### Decisions

- Treat unknown replay signal names as recoverable replay noise.
- If a replayed signal key is not available on the current Vega view/spec, skip applying that signal instead of throwing.
- Keep valid replay signals working normally when they exist on the active spec.
- Prefer a developer-facing warning in analysis/development contexts over an in-app error.
- Add a trial/component boundary guard where practical, but keep the primary safety check in `VegaController` because it is the final boundary before calling Vega.
- Preserve old participant provenance data and replay as much of it as possible; skip only invalid signal keys for the active Vega view.

### Acceptance Criteria

- [ ] Replaying `demo-vega` no longer crashes when a saved `tooltip` event is encountered while the active Vega spec does not define `tooltip`.
- [ ] `VegaController` does not call `view.signal(name, value)` for signal names that are not available on the current Vega view/spec.
- [ ] Valid replay signals, including `revisitAnswer`, `xField`, `yField`, and `hoveredSymbol`, still apply normally when present in the active spec.
- [ ] Existing replay behavior is preserved for Vega specs that do define `tooltip`.
- [ ] A focused unit test covers the unknown-signal replay case.
- [ ] A regression test covers adjacent Vega trials with different signal sets or an equivalent mocked transition.

### Out of Scope

- Rewriting `demo-vega`.
- Migrating or editing existing participant provenance data.
- Changing Vega signal naming conventions for study authors.
- Adding new dependencies.

### Implementation Notes

- The likely failure point is `src/controllers/VegaController.tsx`, where replay currently applies `provState.event.key` directly via `view.signal(...)`.
- `demo-vega` has adjacent Vega trials with different signal sets, which can expose stale or mismatched replay state during transition/loading.
- The defensive check should apply at the controller boundary even if upstream replay state scoping is also tightened.

### Testing Notes

- Add a unit test around `VegaController` that passes a `provState` event with a signal key missing from the mocked/current view and asserts no throw.
- Add coverage that a valid signal still calls `view.signal(...).run()`.
- Use the existing `demo-vega` adjacent-trial shape, or an equivalent mocked pair of Vega specs, for regression coverage.

jaykim1213:
Closes #1268

This PR fixes an occasional `demo-vega` replay crash where Vega throws:
`Uncaught Error: Unrecognized signal name: "tooltip"`

During replay, a signal saved from one Vega chart can sometimes be applied after another Vega chart is shown. If the new chart does not have that signal, Vega throws an error.

This fix checks the signals for the current chart before applying a replayed signal. If the current chart does not have that signal, replay skips it instead of crashing.

### Documentation

- [ ] Done
- [x] N/A

### DIFF
--- a/src/controllers/VegaController.tsx
+++ b/src/controllers/VegaController.tsx
@@ -131,6 +131,8 @@ export function VegaController({ currentConfig, provState }: { currentConfig: Ve
     }, {} as Listeners);
   }, [handleRevisitAnswer, handleSignalEvt, vegaConfig]);
 
+  const configuredSignalNames = useMemo(() => new Set(Object.keys(signalListeners)), [signalListeners]);
+
   useEffect(() => {
     async function fetchVega() {
       setLoading(true);
@@ -154,9 +156,11 @@ export function VegaController({ currentConfig, provState }: { currentConfig: Ve
 
   useEffect(() => {
     if (view && provState && provState.event && provState.event.key) {
-      view!.signal(provState.event.key, structuredClone(provState.event.value)).run();
+      if (configuredSignalNames.has(provState.event.key)) {
+        view.signal(provState.event.key, structuredClone(provState.event.value)).run();
+      }
     }
-  }, [view, provState]);
+  }, [configuredSignalNames, view, provState]);
 
   // If the vega spec can't be fetched (404) or parsed (invalid JSON), clear
   // stimulus validation so the participant isn't stuck on a trial that can
--- a/src/controllers/tests/ComponentController.spec.tsx
+++ b/src/controllers/tests/ComponentController.spec.tsx
@@ -710,4 +710,38 @@ describe('VegaController — signal and event coverage', () => {
 
     expect(mockDispatch).toHaveBeenCalled();
   });
+
+  test('ignores replayed provenance for signals missing from the active vega config', async () => {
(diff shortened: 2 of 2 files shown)


JackWilb (approved):
Reviewed the Vega-only diff against the current dev base. The change safely ignores provenance events for signals missing from the active Vega spec while preserving valid signal replay. Focused unit tests and typechecking pass.

PR #1294, opened 2026-07-09
Save Check Answer progress and fix Enter handling

jaykim1213:

- Saves Check Answer progress, including attempts used and correctness, so it does not reset 
- Moves Enter handling into `NextButton`, so Enter runs Check Answer first when available, then falls back to Next

### Documentation

- [x] Done
- [ ] N/A

### DIFF
--- a/src/components/response/ResponseBlock.tsx
+++ b/src/components/response/ResponseBlock.tsx
@@ -20,7 +20,6 @@ import {
 } from '../../store/store';
 
 import { NextButton } from '../NextButton';
-import { shouldIgnoreEnter } from '../keyboardUtils';
 import {
   generateInitFields, mergeReactiveAnswers, useAnswerField,
 } from './utils';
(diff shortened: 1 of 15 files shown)


src/components/response/ResponseBlock.tsx
```diff
@@ -527,77 +556,34 @@ export function ResponseBlock({
...
-      };
 
-      window.addEventListener('keydown', handleKeyDown);
-      return () => {
-        window.removeEventListener('keydown', handleKeyDown);
-      };
-    }
-    return undefined;
-  }, [checkAnswerProvideFeedback, nextOnEnter, showBtnsInLocation, hasCorrectAnswerFeedback, disabledAttempts]);
+    storeDispatch(setCheckAnswerResult({
```
JackWilb, line 560: **[P1] Persist each grading attempt before Next** — `setCheckAnswerResult` only updates Redux. The only `storageEngine.saveAnswers(...)` call is in `useNextStep`, but Next remains disabled after an incorrect attempt. With `trainingAttempts: 2` and `allowFailedTraining: false`, a participant can answer incorrectly once, refresh, and return with `attemptsUsed` reset to 0; repeating that avoids ever reaching the failure threshold. Persist the updated `checkAnswer` state when grading occurs (or through a storage-backed draft mechanism), and add an incorrect-attempt → refresh/restore test.

JackWilb (changes requested):
Requesting changes because the new Check Answer progress state is only persisted when advancing, so refresh can reset limited training attempts.

src/components/response/ResponseBlock.tsx
jaykim1213: Fixed. Each Check Answer attempt is now saved immediately

JackWilb (approved):
Thanks for addressing the persistence issue. I verified that each Check Answer attempt is now saved immediately, refresh restores both partial and terminal training state, and the focused tests, typecheck, lint, and build pass. I did not find any new actionable issues in the follow-up changes.

PR #1306, opened 2026-07-16
Add managed Trrack provenance APIs

### LINKED ISSUE #659: We need to test with non-linear stimulus history. We are implicitly linear for the form elements.
This will affect the visualization, replay, etc.

JackWilb:
## Summary

- inject a managed `useTrrack` function into every React stimulus through `StimulusParams`, with no reVISit hook import required
- add `Revisit.createTrrack` for website and iframe stimuli
- subscribe before stimulus layout effects, capture the initial node plus every apply, undo, redo, and direct traversal, and clean up listeners on unmount or navigation
- send source-timestamped traversal history through a provenance-only store action instead of coupling provenance writes to answer validation
- keep manual provenance submission and graph-only replay as deprecated compatibility paths for existing studies and historical data
- migrate built-in response and Vega tracking plus all official React, HTML, and Svelte examples to the managed APIs
- regenerate config schemas and add lifecycle, transport, store, replay, and browser-demo coverage

## Root cause

reVISit could only observe graph snapshots that a study explicitly submitted. A private Trrack instance could therefore apply, undo, or redo without the platform seeing the transition, and graph node creation order alone cannot reconstruct revisits to existing nodes.

The managed APIs make reVISit responsible for instance creation and observation while leaving study state and action definitions in study code. React stimuli receive the managed function directly in `StimulusParams`; there is no separate reVISit hook export to import.

## Compatibility

- Existing React stimuli remain compatible because they can ignore the additional `useTrrack` prop.
- Existing `setAnswer({ provenanceGraph })` and `Revisit.postProvenance(...)` integrations continue to work through the store-ingestion fallback and are marked `@deprecated`.
- Historical provenance without traversal events continues through the graph-only replay fallback, which is also marked `@deprecated`.
- No historical timing is fabricated and no dependency is added.

## Validation

- `yarn generate-schemas`
- `yarn typecheck`
- `yarn lint`
- focused Vitest suite: 44 passed
- `yarn unittest --run`: 1,920 passed, 1 skipped
- `yarn build`
- `git diff --check`

The updated Playwright demo coverage will run in CI; it was not run locally per the repository testing policy.

Closes #659


JackWilb:
Can we pass through in StimulusParams?

jaykim1213 (approved):

src/public/tutorial/assets/ReactExample.tsx
```diff
@@ -40,11 +41,10 @@ export default function ReactExample({ parameters, setAnswer, provenanceState }:
     setAnswer({
       answers: {},
       status: true,
-      provenanceGraph: trrack.graph.backend,
     });
   }, [actions, setAnswer, trrack]);
```
jaykim1213, line 21: `ReactExample.tsx` and `BrushPlot.tsx` both call `useTrrack`, so both graphs are saved to the same stimulus provenance field and overwrite each other. 

We should remove Trrack from `ReactExample` and let `BrushPlot` save the provenance data

jaykim1213 (approved):

PR #1311, opened 2026-07-16
Improve study startup assignment performance

### LINKED ISSUE #1176: Lots of studies load really slowly
There are a number of studies where it takes a long time to load them. I know we've fixed slow loading bugs before, but this is omnibus issue for slow loading

JackWilb:
## Summary

This PR improves study-startup performance and sequence-assignment correctness.

It removes participant-history-dependent collection scans from normal startup, eliminates redundant startup reads and large object downloads, and assigns stable sequence and participant indexes using provider-specific concurrency controls. Once the allocator is initialized, the amount of assignment work performed for a new or returning participant is bounded and does not grow with the number of prior participants.

Closes #1176.

## Performance changes

- Replace up to three complete sequence-assignment collection scans during participant startup with participant-specific reads, bounded queries, aggregate counts for legacy records, and allocator state.
- Persist zero-based `sequenceIndex` and `creationIndex` values on new assignments so returning participants can use a point read.
- Reuse the sequence array and modes document already fetched during Shell initialization.
- Use one participant-specific assignment lookup for completion status.
- Skip downloading an unchanged stored config when its active hash already matches.
- Narrow Supabase startup verification from selecting every study row to selecting only the connection row.

The important bound is on client-visible database work: after allocator initialization, normal assignment startup uses a fixed number of point reads, bounded queries, and writes rather than downloading work proportional to participant history. This does not impose a literal wall-clock timeout—authentication, provider latency, retries, and network throughput can still vary—and Firestore transactions may produce more raw HTTP protocol entries even while eliminating redundant logical reads and unbounded document downloads.

### Measured effect

For CALVI with approximately 33 assignments, cache-disabled Firefox HARs showed:

- `dev` downloaded the approximately 3.35 MB sequence array twice per load.
- This PR downloads it once, reducing response bytes from approximately 9.69 MB to 6.41 MB, or about 34%.
- The sequence object was served with `Cache-Control: private, max-age=0`, so the second `dev` download was not expected to become a browser-cache hit when cache was enabled.
- Fast-network first-render timing was effectively unchanged at this study size: median approximately 4.03 seconds on `dev` versus 4.17 seconds on the PR. At roughly 33 assignments, the removed collection scans were inexpensive and the bounded Firebase transaction replaced much of the saved elapsed time.

The immediate benefit is therefore lower transferred data and less duplicate parsing, especially on cold or mobile connections. The database benefit grows with participant history because normal startup no longer downloads the complete assignment collection.

Config-change startup still regenerates and uploads the expanded sequence array; that work is intentionally deferred to the follow-up issues listed below.

## Correctness changes

- Firebase commits the participant assignment, allocator update, and rejected-slot claim in one Firestore transaction. The transaction re-reads the participant, allocator, and candidate rejected assignment so contention causes a retry rather than a duplicate allocation.
- Local storage serializes allocation within the application.
- Supabase uses a versioned compare-and-swap allocator, which prevents duplicate counter reservations under the tested concurrent-start scenario.
- Rejected-slot reuse remains FIFO by assignment timestamp and supports legacy assignments without allocator metadata and chained replacement rejection.
- Existing studies require no assignment migration; allocator state is initialized lazily using provider-side counts.

Automated tests start 12 participants concurrently for Firebase, Supabase, and local storage and verify unique sequence and creation indexes. Provider tests also cover fresh startup, initialized allocators, legacy returning assignments, legacy rejected-slot reuse, oldest rejected-slot selection, chained replacement rejection, and participant-specific completion lookup.

### Correctness caveats

- The Firebase concurrency test uses the repository's serialized in-memory Firestore transaction mock. Rejected-slot behavior is covered functionally, but there is not a dedicated Firebase emulator test in which two simultaneous participants compete for the same rejected slot. The production transaction is designed to handle that race by re-reading and atomically claiming the candidate before commit.
- Supabase allocation is not atomic end to end: rejected-slot claim, allocator reservation, and participant-assignment upsert are separate database operations. The counter reservation is race-safe, but a failure between those operations can leave partial state. #1368 tracks moving the complete operation into an atomic, idempotent database transaction/RPC.

## Firebase deployment requirement

This PR introduces two Firestore compound queries that were not required by the previous full-scan implementation. Firebase's existing automatic single-field indexes, and any unrelated composite indexes already in a project, do not replace them.

Each Firebase project/database must create these collection-scoped composite indexes once:

1. Legacy sequence-position lookup
   - `rejected` ascending
   - `timestamp` ascending
2. Deterministic rejected-slot lookup
   - `rejected` ascending
   - `claimed` ascending
   - `timestamp` ascending

The first index is used only when deriving the position of a legacy assignment without stored allocator metadata. The second query runs during fresh Firebase participant allocation even when no rejected assignment is found, so a fresh Firebase project must create that index before its first participant starts.

These indexes are required once per Firebase project/database, not once per study. They apply to the shared `sequenceAssignment` collection scope and serve dynamically named `prod-{studyId}` and `dev-{studyId}` study collections. If an index already exists in a test project, it may have been created while testing this PR and remains available regardless of which application branch is deployed.

When Firestore reports a missing index, the application shows the exact Firebase-generated creation link, explains that setup is a one-time action for the study owner, and provides open/copy controls.

## Follow-up work

- #1319 moves sequence-array publication off participant startup's critical path.
- #1320 renders the study interface behind a blocking modal while participant startup resolves.
- #1321 replaces expanded sequence arrays with a compact deterministic representation.
- #1322 reduces remaining bounded allocator round trips.
- #1323 profiles and optimizes bundle and authentication startup.
- #1368 makes Supabase sequence allocation atomic end to end.

## Validation

- [x] `yarn unittest --run` — 1,980 passed, 1 skipped
- [x] `yarn typecheck`
- [x] `yarn lint`
- [x] `yarn build`
- [x] Focused provider regression suite — 168 passed
- [x] Local NASA-TLX fresh-load smoke test — 672, 645, 642, 628, 641 ms; median 642 ms
- [x] Firebase fresh and returning-participant HAR captures
- [x] Firebase CALVI before/after HAR captures with browser cache disabled
- [ ] Live Firebase legacy allocator/index query smoke test; the attempted returning-participant test bypassed allocation and therefore did not exercise this query


jaykim1213 (approved):

jaykim1213 (changes requested):
The PR preview currently fails during participant startup because the required sequenceAssignment composite index has not been created in Firebase


Image

jaykim1213 (approved):

PR #1351, opened 2026-07-28
Fix library doc generator and clean up libraries

jaykim1213:
- Fix `libraryDocGenerator.cjs` so custom example study titles stay in the study repo, but the docs still use the library name (e.g. `QUIS: Questionnaire for User Interface Satisfaction` in the study repo, `quis` in the docs)
- Add removing extra whitespaces
- Add removing stale library docs
- Fix library introduction markdown files
- Change references to IEEE style
- Add missing DOIs
- Update UEQ and UEQ-S titles
- Close sidebars for libraries that don't use them
- Add library guide to AGENTS.md
- Fix typos


libraryExampleStudyGenerator.cjs
```diff
@@ -45,7 +47,9 @@ const createExampleConfig = (libraryName) => ({
 });
 
 const getLibraries = (libsPath) => fs.readdirSync(libsPath)
-  .filter(library => !library.startsWith('.') && !library.endsWith('.DS_Store'));
+  .filter((library) => !library.startsWith('.')
+    && !library.endsWith('.DS_Store')
+    && !LIBRARIES_TO_SKIP.has(library));
```
JackWilb: Could we also change the nearby createExampleConfig default to withSidebar: false and assert it in LibraryExampleStudyGenerator.spec.ts? The new guidance says generated examples should keep the sidebar closed unless they actually place content there, but every newly scaffolded library still receives withSidebar: true.

libraryDocGenerator.cjs
```diff
@@ -55,7 +56,20 @@ import StructuredLinks from '@site/src/components/StructuredLinks/StructuredLink
 `;
 
 const getLibraries = (libsPath) => fs.readdirSync(libsPath)
-  .filter((library) => !library.startsWith('.') && !library.endsWith('.DS_Store'));
+  .filter((library) => !library.startsWith('.')
+    && !library.endsWith('.DS_Store')
+    && !LIBRARIES_TO_SKIP.has(library));
```
JackWilb, line 54: Could the generator remove skipped or otherwise stale outputs, or rebuild docsLibraries cleanly? Filtering test here prevents a fresh file from being written, but a pre-existing docsLibraries/test.md survives regeneration and can still be copied to the docs repository with the broken demo link. A regression test that pre-populates the stale file would cover the intended fix.

public/library-adaptive-vlat/assets/adaptive-vlat.md
```diff
@@ -3,7 +3,7 @@
 
 This is a demo of the library `adaptive-vlat`.
 
-A-VLAT is a short, adaptive visualization literacy test, which are built upon the existing static assessments VLAT. 
+A-VLAT is a short, adaptive visualization literacy test, which is built upon the existing static VLAT assessment.
```
JackWilb: This generated line still has trailing whitespace, and the template currently leaves an extra blank line at EOF in 18 generated Markdown files, so git diff --check fails. Could we normalize or trim the generated Markdown, clean the trailing spaces in the source descriptions, regenerate the assets, and add a whitespace assertion to the generator test?

JackWilb (changes requested):
Thanks for the library cleanup and the added generator coverage. I found three generator workflow issues that can preserve or recreate the behavior this PR is trying to fix. I have left concrete suggestions inline.

libraryExampleStudyGenerator.cjs
jaykim1213: Fixed. It now defaults to `withSidebar: false`

libraryDocGenerator.cjs
jaykim1213: Added removing documentation for libraries that are skipped or no longer exists

public/library-adaptive-vlat/assets/adaptive-vlat.md
jaykim1213: Added cleaning whitespaces

JackWilb (approved):
Thanks for addressing the Markdown generation issue. The generator now preserves author-provided formatting while keeping generated section spacing consistent, and the regression coverage verifies the affected fields. The remaining identifier and UEQ-S changes are internally consistent and acceptable for this repository's upgrade cadence.

PR #1357, opened 2026-07-28
Add pairwise ranking validation and disabled state

jaykim1213:
- Pairwise ranking now requires at least one complete pair to proceed
- Ranking widgets are disabled after the answer is finalized 

Screenshot 2026-07-28 at 2 33 37 PM

### DIFF
--- a/src/components/response/RankingInput.tsx
+++ b/src/components/response/RankingInput.tsx
@@ -21,13 +21,16 @@ import cx from 'clsx';
 import {
   Box, Button, Flex, Group, Paper, Stack, Text,
 } from '@mantine/core';
-import { useMemo, useState, useEffect } from 'react';
+import {
+  useCallback, useEffect, useMemo, useState,
+} from 'react';
 import { InputLabel } from './InputLabel';
 import { OptionLabel } from './OptionLabel';
 import classes from './css/RankingDnd.module.css';
 import { ParsedStringOption, RankingResponse, StringOption } from '../../parser/types';
 import { useStoreActions, useStoreDispatch } from '../../store/store';
 import { parseStringOptions } from '../../utils/stringOptions';
+import { getRankingInstanceIndex } from './responseValidation';
 
 type Item = { id: string; symbol: string; option: ParsedStringOption };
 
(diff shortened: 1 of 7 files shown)


src/components/response/responseValidation.ts
```diff
@@ -102,6 +102,25 @@ export function checkNumericalResponse(response: NumericalResponse, value: numbe
...
+  Object.values(value).forEach((location) => {
+    const match = location.match(/^pair-(\d+)-(high|low)$/);
+    if (match) {
+      const [, pairId, position] = match;
+      if (!pairs[pairId]) pairs[pairId] = { high: false, low: false };
+      pairs[pairId][position as 'high' | 'low'] = true;
+    }
+  });
+
+  const hasCompletePair = Object.values(pairs).some((pair) => pair.high && pair.low);
```
JackWilb: The new validator only tracks boolean occupancy, so restored or default data such as `{A_0: "pair-0-high", A_1: "pair-0-low"}` compares an item with itself yet passes. Multiple values in one slot and unknown option IDs also pass. Could we validate exactly one configured option on each side and require the sides to be distinct? Passing `response` into this helper would provide the option set; please add malformed restored/default cases.

src/components/response/RankingInput.tsx
```diff
@@ -470,6 +471,7 @@ function RankingPairwiseComponent({
       <Flex justify="flex-end" mb="md">
         <Button
           variant="outline"
+          disabled={disabled}
```
JackWilb, line 594: Could we initialize `pairIds`, `nextPairId`, and `instanceCounter` from `answer.value` before enabling Add New Pair? On a revisited response containing pair 1 or an existing option instance ending in `_0`, these counters restart at 1 and 0: the first Add click can target an existing pair, and a repeated option can overwrite an existing key. A back-navigation/restoration test would lock this behavior down.

src/components/response/RankingInput.tsx
```diff
@@ -499,6 +501,7 @@ function RankingPairwiseComponent({
                           <SortableItem
                             key={instanceId}
                             item={{ ...item, id: baseItemId, symbol: instanceId }}
+                            disabled={disabled}
```
JackWilb, line 625: We currently detect generated instances with `includes("_")` and recover base IDs using `split("_")[0]`, including just above this line. Structured option values may legally contain underscores; `new_york` is therefore treated as an existing instance and later rendered as `new`, so it disappears while its answer remains stored. Please keep base option IDs separate from generated instance IDs, with an underscore-value test.

JackWilb (changes requested):
Thanks for tackling the pairwise validation and finalized-state behavior. I found three data-integrity issues that should be addressed before merge; details are inline.

src/components/response/responseValidation.ts
jaykim1213: Fixed. It now checks self-pairs, multiple values in one slot, and unknown options. Added tests for malformed restored/default cases

src/components/response/RankingInput.tsx
jaykim1213: Fixed. pairIds, nextPairId, and instanceCounter now initialize from answer.value

src/components/response/RankingInput.tsx
jaykim1213: Fixed. Instead of splitting on `_`, we now look up the real option name from the option list (`getRankingBaseItemId`)

src/components/response/RankingInput.tsx
```diff
@@ -393,8 +440,8 @@ function RankingPairwiseComponent({
       return;
     }
 
-    const isFromUnassigned = !draggedId.includes('_');
-    const baseItemId = isFromUnassigned ? draggedId : draggedId.split('_')[0];
+    const isFromUnassigned = optionIds.has(draggedId);
```
JackWilb: Thanks for the restoration work. There is still an assigned-default case: defaults are documented as plain option-value keys, so an assigned `Pizza` also satisfies `optionIds.has(draggedId)`. Dragging it within the pair enters the from-unassigned branch, writes `Pizza_<counter>`, and leaves the original `Pizza`, duplicating the persisted comparison. Could we derive the source from drag/container metadata or normalize default keys to unambiguous instance IDs? Please add a test that moves an assigned plain default key.

src/components/response/responseValidation.ts
```diff
@@ -102,6 +103,53 @@ export function checkNumericalResponse(response: NumericalResponse, value: numbe
   return null;
 }
 
+export function getRankingBaseItemId(instanceId: string, optionValues: Set<string>): string {
+  if (optionValues.has(instanceId)) {
```
JackWilb, line 121: This exact-match shortcut leaves the option and instance namespaces ambiguous. For example, a Study Designer can configure `{ label: "Original model", value: "model" }` and `{ label: "Model 0", value: "model_0" }`. The first generated instance of `model` is also stored as `model_0`, so this branch resolves the participant’s "Original model" selection as "Model 0" and renders or validates the wrong item. Could we use a tagged or encoded instance-key format that cannot collide with an option value, while retaining legacy-key compatibility? Please add a test covering `model` and `model_0`.

JackWilb (changes requested):
Thanks for addressing the original review. I resolved those superseded threads. Two identifier cases remain before this is safe to merge; concrete reproductions and suggested fixes are inline.

src/components/response/RankingInput.tsx
jaykim1213: Drag ids now include where the item came from (`available`/ `placed`), so moving a placed default key just moves it instead of copying it

src/components/response/responseValidation.ts
jaykim1213: Instance keys are now tagged (`instance-0-model`), so they can't collide with option values like model_0

JackWilb (approved):
Thanks for working through the review. I pushed the remaining ranking edge-case fixes and regression coverage, reran the full local validation gate, and completed a clean adversarial re-review. This is ready from my side.

PR #1390, opened 2026-08-13
Refactor factor syntax and add recursive factor actions

### LINKED ISSUE #842: Additional expressivity in the sequence (factors)
Some sequences are extremely hard to create, despite having straightforward designs. Much of this problem is because there is nothing declarative about the language, you have to do everything. 

We would add 'factors' which can be given designs, such as withinSubjects, and ways to access them within the sequence. This would allow us to cross the factors in the sequence as well, without having to write out the entire crossing in our normal sequence.
```json
  "factors": {
    "visType": { "options": ["nodeLink", "adjMatrix", "map"], "name": "visType", "order": "random", "numSamples": 2, "designType": "betweenSubjects"},
    "dataType": { "options": ["dense", "sparse"], "name": "datasets", "order": "random", "designType": "withinSubjects"},
    "task": { "options": ["search", "neighborhood", "cluster"], "name": "tasks", "samplingStrategy": "linear", "numSamples": 2,  "designType": "betweenSubjects"}
  },
  "sequence": {
    "components" : [
      "intro",
      "consent",
      { "components" : [
          "intro",
          "consent",
          "intro${visType}",
          {
            "factorsToCross": ["visType"],
              "components": [{
                "components": ["training1${visType}", "training2${visType}"],
              },
              {
                "factorsToCross": ["visType", "dataType", "task"],
                "component": "myReactComponent"
              }]
            },
          "post-study survey"
        ]
      }
    ]
  }
```

EdenWuyifan:
## Why

Closes #842.

Factorial studies currently require too much manual enumeration of conditions and generated components. This change gives study designers a recursive factor DSL for describing conditions once, combining them, and placing them in a sequence. It also makes participant-level assignments and runtime sampling explicit so the parser can reject invalid configurations before a study starts.

## Architecture

The parser now separates three responsibilities:

- Factor definitions hold primitive or object-valued levels.
- Recursive expressions combine those levels with `cross`, `zip`, `concat`, `repeat`, `keep`, `remove`, and `sample`.
- Factor blocks compile deterministic expressions into concrete components and retain participant-dependent ordering or sampling as runtime plans.

Between-subject assignments are resolved before runtime factor evaluation. Eligible values are filtered before ordering or sampling, and aliases are translated between the expression input names and the component output parameters. Runtime plans retain their component map and sequence controls so skip, interruption, and conditional behavior continue to be validated and materialized correctly.

Generated factor components are merged from their base components, preserve string-valued template fields, and are checked against the full `IndividualComponent` schema. Invalid between-subject references are parser errors. If participant-specific filtering leaves fewer values than a factor-owned `numSamples` requests, runtime generation stops at the available values and emits a designer-visible warning instead of throwing.

The public study changes add intentional factor fixtures and supporting assets. The factorized configs are in `demo-dsf-study`, `demo-factors`, `demo-ffl-study`, `demo-markdown-factors`, `demo-max-study2`, `demo-stroop-factors`, and `incentives-corr`. `demo-calvi-study` is also added as a study fixture, but it does not use the factor DSL.

## User-facing impact

Study designers can express reusable factorial conditions and participant assignments directly in config files. Invalid factor references now fail during parsing, while valid studies can generate participant-specific trials without manually enumerating every component. The added public studies provide examples and assets for these configurations.

## Validation

- `corepack yarn unittest` — 2,280 tests passed, 1 skipped.
- `corepack yarn typecheck` — passed.
- `corepack yarn lint` — passed with one pre-existing `StepsPanel.tsx:599` hook-dependency warning.
- `corepack yarn build` — passed.
- `node scripts/check-lockfile-drift.mjs` and `git diff --check` — passed.
- The previous GitHub Actions run for commit `36a69cd6` passed lint, deploy, unit tests, all Chromium shards, and WebKit shards 1, 3, and 4; WebKit shard 2 failed. The newly pushed commit `211a5cac` has not reported checks yet.

## Risk and rollout

No migration or deployment sequencing is required. The main compatibility change is stricter parser validation for malformed factor configs. A known follow-up remains: analysis/statistics still enumerate canonical `studyConfig.components`, so participant-specific factor-generated component reporting needs a separate change. The previous run had a WebKit shard 2 failure; the new head is awaiting checks.


ZachCutler04:
## Update: factor API work added since this PR opened

The config language has been changed a bit since the PR was opened. The original PR description still provides background; this comment records the current API and implementation changes.

### Config and type changes

- **Objects allowed in factor definitions are supported.** A factor can now be a js object, not just a primitive. This is to support keeping multiple data parameters grouped together, and will be a fairly rare use case. Nested objects are not supported well. 
- **`betweenSubjects` accepts derived factors too.** Between subjects works with derived factors, the same way it did previously with non-derived factors. 
- **Factors can own their ordering and sampling:**
  ```json
  "stimuli": {
    "values": [/* stuff */],
    "order": "random",
    "numSamples": 5
  }
  ```
  This makes the participant-specific order an attribute of the factor, shared wherever that factor is referenced, instead of scattering ordering logic across sequence blocks. This is important for some studies where we want to make derived factors based on the ordering of one factor - requiring the randomization be done first, in the original factor definition.
- **`cross` and `zip` accept `as`** to name their output parameters explicitly. This avoids positional generated names and makes expressions reusable, so you can cross a factor with itself.
- **`keep` and `remove` are added as operators functioning as filters Both operate on a source `factor` and select conditions either by partial `condition` matching or explicit `items`.
- **`sample` now requires an explicit `samplingStrategy`:** `withoutReplacement` or `withReplacement`. The former is used for distinct assigned conditions; the latter intentionally permits repeats.
- **Sequence blocks no longer accept ad-hoc `parameters`.** Parameters are declared on components/base components and filled from factors. This prevents the sequence from becoming a second place to define factor values.


The branch now includes factorized demos based on several real study designs (max's guardrails, FFL, data-serialization formats, CALVI, and visualization-complexity ratings). They exercise coupled object conditions, between-subject objects, `keep`/`remove`, explicit sampling strategies, zipped/crossed assignment, and large ordered stimulus pools.

JackWilb:
@codex review

JackWilb (approved):

PR #1392, opened 2026-08-14
disable audio/screen recording if datacolleciton is disabled

### LINKED ISSUE #1183: Disable audio / screen capture recording when data collection is turned off
Right now, even when data collection is turned off, the microphone is on. Not sure whether data is stored or not. 

We should really turn that off. 

There will be some work to do also on the test widgets for this, they might need a special mode for data collection turned off that lets ppl pass in all cases if that's the situation.

A7700:
Closes #1183 .

When data collection is turned off the audio recording seemed to still be happening, now we escape before recording or any visual indication for it is triggered and we can proceed without needing to use the recording.  In the case of screen recording you couldn't proceed because of errors and recording had to be started, now the start recording button is disabled so no accidental data collection is started. If data collection is enabled the behavior should be the same as before.

Also there is a small hotfix: when the permissions for the audio is not given but instead the site is reloaded the user could bypass the required answer and continue without audio, now there will be an error if audio is required and no permission has been granted as it should be.

Before:
image
image

Now:
image
image

### DIFF
--- a/src/public/libraries/screen-recording/assets/ScreenRecording.tsx
+++ b/src/public/libraries/screen-recording/assets/ScreenRecording.tsx
@@ -5,6 +5,7 @@ import { useEffect, useState } from 'react';
 import { useRecordingContext } from '../../../../store/hooks/useRecording';
 import { StimulusParams } from '../../../../store/types';
 import { RecordingAudioWaveform } from '../../../../components/interface/RecordingAudioWaveform';
+import { useStoreSelector } from "../../../../store/store";
 
 function ScreenRecordingPermission({ setAnswer }: StimulusParams<undefined>) {
   const {
@@ -19,8 +20,18 @@ function ScreenRecordingPermission({ setAnswer }: StimulusParams<undefined>) {
 
   // audioCapturingSuccess is set to true when we detect sound.
   const [audioCapturingSuccess, setAudioCapturingSuccess] = useState(false);
+  const { dataCollectionEnabled } = useStoreSelector((state) => state.modes);
 
   useEffect(() => {
+    if (!dataCollectionEnabled) {
+      setAnswer({
+        status: true,
+        answers: {
+          audioTest: true,
+        },
+      });
+      return;
+    }
     setAnswer({
       status: screenCapturing && (studyHasAudioRecording ? audioCapturingSuccess : true),
       answers: {
@@ -102,7 +113,10 @@ function ScreenRecordingPermission({ setAnswer }: StimulusParams<undefined>) {
               <strong>Click the button below</strong>
               {' '}
               to enable screen and audio recording.
-              <Button type="button" onClick={screenCapturing ? stopCapture : startCapture} display="block" mt="sm">
+              <Button type="button"
+                      onClick={screenCapturing ? stopCapture : startCapture}
+                      disabled={!dataCollectionEnabled}
+                      display="block" mt="sm">
                 {screenCapturing ? 'Stop Recording' : 'Start Recording'}
               </Button>
               <video
@@ -146,7 +160,10 @@ function ScreenRecordingPermission({ setAnswer }: StimulusParams<undefined>) {
           <strong>Click the button below</strong>
           {' '}
           to enable screen recording.
-          <Button type="button" onClick={screenCapturing ? stopCapture : startCapture} display="block" mt="sm">
+          <Button type="button"
+                  onClick={screenCapturing ? stopCapture : startCapture}
+                  disabled={!dataCollectionEnabled}
+                  display="block" mt="sm">
             {screenCapturing ? 'Stop Recording' : 'Start Recording'}
           </Button>
           <video
--- a/src/public/libraries/screen-recording/assets/ScreenRecording.tsx
+++ b/src/public/libraries/screen-recording/assets/ScreenRecording.tsx
@@ -38,7 +38,7 @@ function ScreenRecordingPermission({ setAnswer }: StimulusParams<undefined>) {
         screenRecordingPermission: screenCapturing,
       },
     });
-  }, [screenCapturing, audioCapturingSuccess, setAnswer, studyHasAudioRecording]);
+  }, [screenCapturing, audioCapturingSuccess, setAnswer, studyHasAudioRecording, dataCollectionEnabled]);
 
   useEffect(() => {
     if (!screenCapturing) {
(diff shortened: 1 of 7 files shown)


src/store/hooks/useRecording.ts
```diff
@@ -40,6 +41,7 @@ export function useRecording() {
   const [isSpeakingWhileMuted, setIsSpeakingWhileMuted] = useState(false);
   const [analysisStreamReady, setAnalysisStreamReady] = useState(false);
   const [showMutedWarning, setShowMutedWarning] = useState(false);
+  const modes = useStoreSelector((state) => state.modes);
```
JackWilb, line 44: Thanks for making the data-collection gate explicit here. This new Redux selector causes the existing `useRecording` tests—and the analogous waveform tests—to fail because they render without a Redux provider. Could you update the test wrappers/mocks and add coverage asserting that disabled collection makes zero media API, recorder, and storage calls? The Unit Tests job currently reports 32 failures from this provider issue.

src/public/libraries/screen-recording/assets/ScreenRecording.tsx
```diff
@@ -19,8 +20,18 @@ function ScreenRecordingPermission({ setAnswer }: StimulusParams<undefined>) {
...
   // audioCapturingSuccess is set to true when we detect sound.
   const [audioCapturingSuccess, setAudioCapturingSuccess] = useState(false);
+  const { dataCollectionEnabled } = useStoreSelector((state) => state.modes);
 
   useEffect(() => {
+    if (!dataCollectionEnabled) {
+      setAnswer({
+        status: true,
+        answers: {
+          audioTest: true,
```
JackWilb: Could this use `screenRecordingPermission: true` instead? That is the response key declared by the screen-recording library. Using `audioTest` lets the stimulus pass, but can make skip conditions and collected trial answers diverge when data collection is disabled.

src/store/store.tsx
```diff
@@ -482,6 +482,11 @@ export function useAreResponsesValid(id: string) {
     if (id.includes('reviewer-')) {
       return true;
     }
+
+    if (!state.trialValidation[id]) {
```
JackWilb: Could we remove this guard for now? It changes the established behavior for IDs without validation, and the existing unit test now fails. I also couldn’t find production callers of `useAreResponsesValid`, so this does not appear to fix the reload path; the `AudioTest` invalidation is the targeted fix. A focused reload/denied-permission regression test would protect that behavior better.

JackWilb (changes requested):
Thanks for taking on issue #1183 and for your first contribution to ReVISit! The overall direction is useful: media acquisition and recording should be gated when data collection is disabled, and the microphone check should not be bypassable after reload.

I’m requesting a few changes before merge:

- Make the Unit Tests pass and add disabled-mode regression coverage.
- Emit `screenRecordingPermission` from the screen-recording component.
- Remove or revisit the unused validation guard.

Lint/build, Chromium, WebKit, and the preview deployment are passing. Once these changes are addressed, I’ll be happy to re-review.

src/public/libraries/screen-recording/assets/ScreenRecording.tsx
```diff
@@ -19,8 +20,18 @@ function ScreenRecordingPermission({ setAnswer }: StimulusParams<undefined>) {
 
   // audioCapturingSuccess is set to true when we detect sound.
   const [audioCapturingSuccess, setAudioCapturingSuccess] = useState(false);
+  const { dataCollectionEnabled } = useStoreSelector((state) => state.modes);
```
JackWilb, line 23: Thanks for adding the mode check here. Could we also apply it to the combined screen-and-audio Start button and cover both button branches in a component test? The screen-only button is disabled when collection is off, but the combined button remains clickable. Clicking it can enter `startScreenCapture` and seed the later rejection state even though no media is captured. Please add `disabled={!dataCollectionEnabled}` to the combined button and verify that both disabled-mode variants remain non-interactive.

src/store/hooks/useRecording.ts
```diff
@@ -355,7 +366,7 @@ export function useRecording() {
         setRecordingError(null);
         setAudioRecordingError(null);
 
-        const screenStream = studyHasScreenRecording ? await navigator.mediaDevices.getDisplayMedia({
+        const screenStream = (studyHasScreenRecording && dataCollectionEnabled) ? await navigator.mediaDevices.getDisplayMedia({
```
JackWilb: Could we make disabled collection an immediate no-op at the beginning of `startScreenCapture`? These conditions prevent the media permission calls, but the function continues, creates an empty stream, and sets `screenCaptureStarted` to true. After navigation, that causes the rejection effect to fire. An early `if (!dataCollectionEnabled) return;` before changing the title or state would fix the behavior and let us remove the redundant mode checks from the individual media calls.

src/store/hooks/tests/useRecording.spec.ts
```diff
@@ -446,3 +455,50 @@ describe('useRecordingContext', () => {
...
+
+    const { result } = renderHook(() => useRecording());
+    await act(async () => { /* let effects settle */ });
+
+    expect(navigator.mediaDevices.getUserMedia).not.toHaveBeenCalled();
+    expect(mockStorageEngine.saveAudioRecording).not.toHaveBeenCalled();
+    expect(result.current.isAudioRecording).toBe(false);
+  });
+
+  test('does not call getUserMedia/getDisplayMedia/recorder/storage via startScreenCapture', async () => {
```
JackWilb, line 506: Could this test verify the complete disabled-mode behavior? It currently does not assert `getDisplayMedia` or `MediaRecorder`, and it passes even though `isRejected` becomes true for the mocked `intro` component. Please assert both media APIs and recorder construction are untouched, all capture flags remain false, and rerendering on the next component does not trigger rejection. That would catch the current regression rather than only checking storage calls.

JackWilb (changes requested):
Thanks for the thoughtful follow-up. The earlier unit-test failures, response key, and unused validation guard are resolved, and the current checks are green. I found one remaining disabled-mode progression issue: the combined screen-and-audio control can still enter a false “capture started” state and reject the participant after navigation.

I’m requesting changes to make disabled screen capture an immediate no-op, disable the combined-mode button, and extend the regression coverage to include capture state, post-navigation rejection, and the microphone reload/denial path. Once those are addressed, I’ll be happy to re-review.

JackWilb (approved):
Thanks for addressing the recording and disabled-data-collection feedback. The requested changes are resolved on this head, and the recording behavior and focused tests look good. The remaining WebKit ranking-widget failure is unrelated to this pull request and is being addressed on a separate branch.

PR #1395, opened 2026-08-19
Added className to asteriskNode in MarkdownWrapper

AF-Cabouat:
I didn't open an issue about it.

Currenyl to target the "required" asterisk through css
- when there are lots of questions, allows to make the mark less distracting (e.g., blue instead of read)
- when all questions are mandatory, allows to hide the node


src/components/ReactMarkdownWrapper.tsx
```diff
@@ -77,7 +77,7 @@ export function ReactMarkdownWrapper({ text, required, inline }: { text: string;
       const asteriskNode: Element = {
         type: 'element',
         tagName: 'span',
-        properties: { style: 'color: #fa5252; margin-left: 4px' },
+        properties: { className: 'required-asterisk', 'style: 'color: #fa5252; margin-left: 4px' },
```
JackWilb: Thanks for adding this customization hook - the stable selector makes sense here. The immediate parse fix is this exact line:

```ts
properties: { className: 'required-asterisk', style: 'color: #fa5252; margin-left: 4px' },
```

Here, `className` and `style` are unquoted property keys, while their values are quoted strings. This exact object literal parses correctly.

Separately, keeping `color: #fa5252` inline means ordinary study CSS such as `.required-asterisk { color: blue; }` will still render red unless it uses `!important`. Could you move the default color into an overridable global rule while retaining this stable class, or explicitly document and test the `!important` requirement if that is intentional?

JackWilb (changes requested):
Thanks for contributing this—adding a stable CSS hook is a useful, focused improvement. I’m requesting a few small changes before this is ready:

- Could you retarget this PR from `main` to `dev`? The repository’s documented flow is feature branch → `dev` → release PR → `main`, and merging directly to `main` would trigger the production deployment.
- After retargeting, could you extend the existing `ReactMarkdownWrapper` required-field test on `dev` to assert that `.required-asterisk` is present? That protects the selector itself rather than only checking that an asterisk is rendered.

The inline comment covers the parsing error and the inline-color behavior. Once those are addressed, the change can stay nicely concise—there’s no need for a broader refactor, new configuration, or dependency. Thank you!

src/components/ReactMarkdownWrapper.tsx
AF-Cabouat: Thank you for the review and very sorry for my typing error -- I'm editing in the browser.
I'm also not very familiar with PR and I reverted my initial commit before pushing the changes so I'll have to make a new PR. 
Sorry again for the mess!

PR #1403, opened 2026-08-20
Add Handlebars templating support and demo-templating study

### LINKED ISSUE #1142: String templates for config files
It would be nice if we can introduce some sense of "dynamic data binding" to the configuration files. 

Some kind of delimiter (maybe double brackets? `{{...}}` or something that starts with a dollar sign? `${...}`) 

Ideally something that allows me to make the following specification (under `baseComponents`):  
```json
"training-trial": {
    "type": "react-component",
    "path": "my-study/assets/training.tsx",
     "parameters": {
        "train_val": 20
      }
    "response": [{
        "id": "training-ans",
        "prompt": "What is area under the curve for values left of {{parameters.train_val}}?",
        "required": true,
        "type": "numerical"
    }],
    "instructionLocation": "aboveStimulus"
},
```

**Describe alternatives you've considered**
Haven't thought about alternatives other than writing my own React components that captures input data, but then it would be poorly integrated with trrack ... 

**Additional context**
N/A

### LINKED ISSUE #1148: Injecting variables into revisit components
In the experiment I am implementing, I want the question text to vary based on the stimulus shown to participants. I have a [base component](https://github.com/abhsarma/adversarial-effects-experiment/blob/c025ac084746a7b9f9cb1c8fb64b83024c93eb24/src/public/incentives-dm/assets/03-training.tsx#L13), and I am passing data using `parameters` which changes the stimuli, but I am using Revisit's response interface which I am declaring in the [config file](https://github.com/abhsarma/adversarial-effects-experiment/blob/main/public/incentives-dm/config.json).

I was trying to inject the relevant text using JS query selector / d3.select, but since revisit re-render's the components, the text disappears on every interaction (click, keydown etc.). Is there a simple way to do this, short of creating my own response form?

Ideally, this would be solved by allowing variable declaration in the config file but I am wondering if there is the possibility of a quick short-term fix which is less hack-y then the query selector approach.

### LINKED ISSUE #1385: Variable Injection / Templates in form elements
Related to #1142, every time we specify text for a form element (aka response), it should also work with a template / variable.

### LINKED ISSUE #1388: Get data from previous questions
By @abhsarma 

Qualtrics allows responses to previous questions to be references in subsequent questions using [piped text](https://www.qualtrics.com/support/survey-platform/survey-module/editing-questions/piped-text/piped-text-overview/#PipingFromAPreviousQuestion).
I imagine a version of this being implemented in ReVISit for some questions which have explicit labels. 
This can be subsequently extended to even questions which are generated for factor components by having such questions indexed in a principled way.

Related and possibly solvable by #1027

### LINKED ISSUE #1398: String templating
This issue brings together a few disparate issues, including #1142, #1148, #1385 

We want to use string templates with a double curly brackets syntax {{variable}}

We want to use string templates in three contexts: 

1. In the revisit spec, so that it can be applied to 
  * form elements, including text in instructions. For example, "Did you enjoy interacting with this {{visualization_name}}?" in a question. visualization_name is a parameter that's available to the component. 
  * for factors to specify components. For example {{vistechnique}}.html would replace to different files. 
2. In Markdown files, so that it can be used to partially adjust text (blocks)
3. in HTML files, with the same purpose as 2.

hilsonshrestha:
Support {{variable}} placeholders, resolved from a component's parameters object, in instruction text, response prompt/secondaryText/infoText, help text (helpTextPath), and per-controller paths and fetched content (markdown, image, video, iframe, vega, react-component).

Adds Handlebars helpers on top of plain substitution:
- `ifEquals` (with `else ifEquals`/`else` chaining) for conditional branches
- `lookupAnswersRel` for reading another trial's answer relative to the current one (e.g. `-1` = previous trial)
- `lookupAnswers` for reading another trial's answer by absolute position in the sequence, with Python-style negative indexing (`-1` = last trial)

Adds a demo-templating study that reuses a single base component across six capital-cities trials to showcase substitution and both lookup helpers across instruction, hint markdown, help text, and response prompt/secondaryText/infoText.

Documentation:
https://github.com/revisit-studies/reVISit-studies.github.io/pull/264

Closes #1398 #1142 #1148 #1385


src/utils/handlebars.ts
```diff
@@ -0,0 +1,55 @@
...
+  'lookupAnswers',
+  (index: number, responseId: string, options: Handlebars.HelperOptions) => {
+    const { answers, flatSequence } = (options.data ?? {}) as {
+      answers?: Record<string, StoredAnswer>; flatSequence?: string[];
+    };
+    if (!answers || !flatSequence) {
+      return undefined;
+    }
+    // Python-style negative indexing: -1 is the last step, -2 the second-to-last, etc.
+    const resolvedIndex = index < 0 ? flatSequence.length + index : index;
```
JackWilb: Please index an answer-bearing sequence rather than flatSequence directly. Runtime participant sequences append an end sentinel, so -1 resolves to end_<index>, which has no stored answer. The new test fixture omits end and therefore passes while production behavior fails. Please add a runtime-shaped regression test and define handling for dynamic-block identifiers.

src/utils/handlebars.ts
```diff
@@ -0,0 +1,55 @@
...
+      return undefined;
+    }
+    const identifier = `${flatSequence[resolvedIndex]}_${resolvedIndex}`;
+    return answers[identifier]?.answer?.[responseId];
+  },
+);
+
+export function compileTemplate(text: string, parameters: Record<string, unknown> = {}, options?: { noEscape?: boolean; data?: Record<string, unknown> }): string {
+  try {
+    return Handlebars.compile(text, { noEscape: options?.noEscape })({ ...parameters, REVISIT: options?.data }, { data: options?.data });
```
JackWilb, line 113: Please expose the parameters object under parameters as well as keeping the root aliases. Linked issue #1142 uses the exact {{parameters.train_val}} form; this context currently only exposes train_val and REVISIT, so that documented/requested syntax renders empty.

src/utils/tests/handlebars.spec.ts
```diff
@@ -0,0 +1,143 @@
...
+      trial_1: makeAnswer('trial-answer'),
+      outro_2: makeAnswer('outro-answer'),
+    },
+  };
+
+  test('looks up an answer by absolute (positive) index', () => {
+    expect(compileTemplate('{{lookupAnswers 1 "response"}}', {}, { data })).toBe('trial-answer');
+  });
+
+  test('supports Python-style negative indexing (-1 is the last step)', () => {
```
JackWilb, line 92: This new regression test uses flatSequence without the runtime end sentinel, so it does not test the behavior users will receive. Please include the real sequence shape (and an answer-bearing last trial), plus coverage for {{parameters.foo}} and dynamic identifiers.

src/components/interface/HelpModal.tsx
```diff
@@ -25,6 +27,13 @@ export function HelpModal() {
 
   const helpTextPath = useMemo(() => componentConfig.helpTextPath ?? config.uiConfig.helpTextPath, [componentConfig.helpTextPath, config.uiConfig.helpTextPath]);
 
+  const templateData = useTemplateAnswerContext();
+
+  const templatedHelpText = useMemo(
```
JackWilb, line 53: This compiles fetched help content, but helpTextPath itself is still passed raw to the fetch below. A path such as help-{{condition}}.md cannot load. Please compile the selected component/global path with the same parameters and answer context before fetching, and use that resolved path for errors.

src/controllers/ImageController.tsx
```diff
@@ -6,14 +6,20 @@ import { ImageComponent } from '../parser/types';
 import { PREFIX } from '../utils/Prefix';
 import { getStaticAssetByPath } from '../utils/getStaticAsset';
 import { ResourceNotFound } from '../ResourceNotFound';
+import { compileTemplate } from '../utils/handlebars';
 
 export function ImageController({ currentConfig }: { currentConfig: ImageComponent; }) {
+  const templatedPath = useMemo(
+    () => compileTemplate(currentConfig.path, currentConfig.parameters ?? {}, { noEscape: true }),
```
JackWilb: This path compilation receives no data, so lookupAnswers and lookupAnswersRel silently resolve empty here (and the same pattern exists in the iframe, video, Vega, and React-component controllers). Please pass the template answer context consistently, or explicitly restrict these helpers to text/Markdown contexts.

src/components/interface/AppNavBar.tsx
```diff
@@ -29,8 +31,12 @@ export function AppNavBar({
   }, [stepConfig, studyConfig]);
 
   const status = useStoredAnswer();
+  const templateData = useTemplateAnswerContext();
 
-  const instruction = currentConfig?.instruction || '';
+  const instruction = useMemo(
```
JackWilb, line 46: This compiles only the static inherited config parameters. Dynamic components store their generated parameters in useStoredAnswer(), and ComponentController already overlays those values for the main surface; the sidebar instruction/response path does not. Please overlay the stored answer parameters here, and apply the same rule to help text.

src/controllers/IframeController.tsx
```diff
@@ -138,9 +144,9 @@ export function IframeController({ currentConfig, provState, answers }: { curren
         pointerEvents: isAnalysis ? 'none' : undefined,
       }}
       src={
-        currentConfig.path.startsWith('http')
-          ? currentConfig.path
-          : `${BASE_PREFIX}${currentConfig.path}?trialid=${currentComponent}&id=${iframeId}`
+        templatedPath.startsWith('http')
```
JackWilb, line 156: Templating the iframe URL does not template placeholders inside a local HTML file, which remains loaded unchanged here. That leaves the HTML-content requirement from linked issue #1398 unmet. Please compile same-origin HTML safely (preserving relative assets and ReVISIT communication), or narrow the scope and issue claim.

package.json
```diff
@@ -58,6 +58,7 @@
     "diff": "^8.0.2",
     "eslint-config-airbnb": "^19.0.4",
     "firebase": "^11.2.0",
+    "handlebars": "^4.7.9",
```
JackWilb, line 61: Please update the tracked package-lock.json with this dependency. The current required lockfile-drift check fails and generates missing Handlebars/transitive entries.

public/demo-templating/config.json
```diff
@@ -0,0 +1,152 @@
+{
+  "$schema": "https://raw.githubusercontent.com/revisit-studies/study/v2.4.3/src/parser/StudyConfigSchema.json",
```
JackWilb, line 2: This demo points at the released v2.4.3 schema, which predates parameters on these component types. Editors therefore report schema errors even though the PR adds a newer local schema. Point to a compatible released schema or defer the demo until the schema is released.

public/demo-html/assets/introduction.md
```diff
@@ -1,8 +1,13 @@
...
+
+- [The Config](https://github.com/revisit-studies/study/blob/main/public/demo-html/config.json)
+- [bar-chart.html](https://github.com/revisit-studies/study/blob/main/public/demo-html/assets/bar-chart.html)
 
 ## Relevant documentation:
- * [Designing HTML Stimuli](https://revisit.dev/docs/designing-studies/html-stimulus/)
+
+- [Designing HTML Stimuli](https://revisit.dev/docs/designing-studies/html-stimulus/)
+
+hello {{n}}
```
JackWilb: Please remove these unrelated debug lines from the existing HTML demo. They render an unresolved {{n}} and test markup in a study with no parameter, while the new templating demo already provides the intended example.

public/demo-templating/assets/quiz-help.md
```diff
@@ -0,0 +1,18 @@
...
+On the previous question you answered **"{{lookupAnswersRel -1 'capital-answer'}}"**.
+{{else}}
+This is your first capital-city question — there's no previous answer to reference yet.
+{{/if}}
+
+All of your answers in one place:
+|Tasks|Identifier|Answers|
+|-----|----------|-------|
+{{#each REVISIT.answers}}
+|TASK {{@index}}|{{this.identifier}}|{{lookupAnswers @index 'capital-answer'}}|
```
JackWilb: Please remove the trailing space after the final |; git diff --check currently reports this line.

JackWilb (changes requested):
Thanks for the follow-up commits—the focused utility tests and duplicate-trial cleanup are useful. I re-reviewed the new head and the complete change set. Request changes remains necessary for the following confirmed issues:

- Answer lookup indexes the runtime `end` sentinel and cannot resolve dynamic-block answer identifiers; the new `-1` test omits `end` and masks this.
- The linked {{parameters.foo}} syntax is still unsupported.
- helpTextPath is not templated, and answer helpers are unavailable in most controller path substitutions.
- Runtime-generated parameters are not applied in the sidebar/help surfaces.
- HTML-file content templating from #1398 is not implemented.
- package-lock.json is stale and the required lockfile-drift check fails.
- The demo still contains unrelated debug content, points at an incompatible released schema, and has diff-check formatting errors.

The new unit suite passes, but it does not exercise real sequences containing `end`, dynamic blocks, rendered controller/help paths, or the demo in Chromium. Please address the inline comments and rerun the focused and required checks.

src/utils/handlebars.ts
hilsonshrestha: Decided with {{parameter_value}} directly instead of {{parameters.value}} for simplicity

src/controllers/IframeController.tsx
hilsonshrestha: We won't be doing HTML templating at the moment. This can be done using revisit communicate js instead.

src/controllers/ReactComponentController.tsx
```diff
@@ -20,7 +22,9 @@ const modules = import.meta.glob(
 ) as Record<string, ModuleNamespace>;
 
 export function ReactComponentController({ currentConfig, provState, answers }: { currentConfig: ReactComponent; provState?: unknown, answers: ParticipantData['answers'] }) {
-  const reactPath = `../public/${currentConfig.path}`;
+  const templateData = useTemplateAnswerContext();
+  const templatedPath = compileTemplate(currentConfig.path, currentConfig.parameters ?? {}, { noEscape: true, data: templateData });
+  const reactPath = `../public/${templatedPath}`;
```
ZachCutler04: (slop but seems right) react-component paths are intentionally limited to static, config-resolvable paths. While other asset paths may be templated at runtime, React components must be discoverable by Vite’s build-time module map and validated by the parser.
We should explicitly document that Handlebars expressions—including {{parameters}} and answer-lookup helpers—are unsupported in react-component.path, and update the parser error to explain this constraint.

src/utils/tests/handlebars.spec.ts
```diff
@@ -0,0 +1,202 @@
+import {
+  describe, expect, test, vi,
+} from 'vitest';
+import { compileTemplate } from '../handlebars';
+import { StoredAnswer } from '../../store/types';
+
```
ZachCutler04, line 6: This testing looks good, but we should probably also add some playwright tests for this stuff? That shouldnt be too bad?

ZachCutler04 (changes requested):

src/controllers/ReactComponentController.tsx
hilsonshrestha: reactPath with template works. I have skipped file existence check for path containing `{{` (template placeholder)

src/components/interface/AppNavBar.tsx
```diff
@@ -29,8 +31,12 @@ export function AppNavBar({
...
+    () => compileTemplate(currentConfig?.instruction || '', currentConfig?.parameters ?? {}, { data: templateData }),
+    [currentConfig?.instruction, currentConfig?.parameters, templateData],
```
JackWilb: Dynamic components store their generated parameters in answers[currentIdentifier].parameters, and ComponentController overlays those values before rendering the main surface. This sidebar path still compiles only currentConfig.parameters, so templated sidebar instructions (and the ResponseBlock rendered below) will use empty/stale values for dynamic trials. Please reuse the same runtime-parameter merge here and in the help surface, with a rendered dynamic-block regression test.

src/store/hooks/useTemplateAnswerContext.ts
```diff
@@ -0,0 +1,27 @@
...
+// answers. currentComponent/funcIndex are only non-trivial inside a dynamic block, where a
+// single flatSequence step fans out into many answer-bearing iterations; without them,
+// `lookupAnswersRel` can't tell which iteration is "current" and relative lookups would be
+// ambiguous.
+export function useTemplateAnswerContext() {
+  const answers = useStoreSelector((state) => state.answers);
+  const flatSequence = useFlatSequence();
+  const currentStep = useCurrentStep();
+  const currentComponent = useCurrentComponent();
+  const { funcIndex } = useParams();
```
JackWilb, line 26: This hook creates a second stateful useCurrentComponent() instance inside each controller. For a dynamic block that resolver begins as __dynamicLoading and is corrected in an effect; the controller can therefore compile an answer-dependent path once with the wrong component/current iteration and start an incorrect fetch before the corrected render. Please defer template/path evaluation until the dynamic component is resolved, or pass the already-resolved component context down from the parent.

src/utils/handlebars.ts
```diff
@@ -0,0 +1,109 @@
...
+    }
+    const staticIdentifier = `${componentName}_${index}`;
+    if (staticIdentifier in answers) {
+      identifiers.push(staticIdentifier);
+      return;
+    }
+    const dynamicPrefix = `${componentName}_${index}_`;
+    Object.entries(answers)
+      .filter(([key]) => key.startsWith(dynamicPrefix))
+      .sort(([, a], [, b]) => (parseTrialOrder(a.trialOrder).funcIndex ?? 0) - (parseTrialOrder(b.trialOrder).funcIndex ?? 0))
```
JackWilb, line 42: This broad prefix can absorb a regular answer from a later component whose name starts with the same string. For example, a dynamic block named block at step 1 will also match a regular component named block_1_trial at step 2 (block_1_trial_2), then count that answer again at its real step. Please validate the matched answer’s parsed trial step/identifier (and deduplicate) before treating it as a dynamic iteration; add a collision regression test.

src/parser/parser.ts
```diff
@@ -132,6 +132,9 @@ function verifyReactComponent(
     'path' in component
       && component.path != null
       && component.type === 'react-component'
+      // A templated path (e.g. `{{file}}.tsx`) can't be resolved until runtime, once
+      // parameters/answers are known, so it can never match a real file in this static glob.
+      && !component.path.includes('{{')
       && !(`../public/${component.path}` in modules)
```
JackWilb, line 169: This treats any {{ substring as a valid deferred template, so malformed paths such as foo{{bar bypass static validation. compileTemplate then catches the syntax error at runtime and returns the raw unresolved path, leaving the invalid config accepted until render time. Please validate the Handlebars syntax before suppressing the existence check, or document/test this fallback explicitly.

JackWilb (changes requested):
Thanks for the follow-up lockfile fix. I rechecked the current head and am requesting changes for four remaining issues:
- Dynamic runtime parameters are not merged into sidebar/help surfaces.
- Dynamic template contexts can begin with __dynamicLoading, causing an incorrect first path evaluation/fetch.
- Dynamic answer matching can collide with regular component identifiers.
- Malformed React-component templates bypass parser validation.
The HTML-content item is intentionally omitted from this review as agreed.

src/parser/parser.ts
hilsonshrestha: Should we remove support for react files like Zach suggested?

hilsonshrestha: Good catch — switched to a regex that requires balanced `{{ ... }}` expressions (`^([^{}]|\{\{[^{}]+\}\})*$`), so a stray `{{` like `foo{{bar` no longer counts as templated and still gets caught by the existence check. Added a test for the malformed case.

src/store/hooks/useTemplateAnswerContext.ts
hilsonshrestha: Went with your suggested approach: `useTemplateAnswerContext` now returns `undefined` while its own `useCurrentComponent()` instance is still `'__dynamicLoading'`, instead of handing back stale/wrong-iteration data. Every consumer (ReactComponentController, VegaController, IframeController, ImageController, VideoController, MarkdownController, AppNavBar, HelpModal, ResponseSwitcher) now skips path/template compilation and asset fetching until it resolves.

Added `useTemplateAnswerContext.spec.ts` proving the hook returns `undefined` during `__dynamicLoading` and resolves correctly on a later render, plus new IframeController tests proving no iframe (and therefore no request) is rendered while `__dynamicLoading`, and that it renders with the correct path once resolved.

hilsonshrestha: Apologies — the fix (commit 937b81f6) was committed locally but hadn't actually been pushed yet when I replied, so the PR was still showing the unfixed version. It's pushed now: `useTemplateAnswerContext` returns `undefined` while its own `useCurrentComponent()` is `'__dynamicLoading'`, and every consumer skips compilation/fetching on that undefined case. Sorry for the confusion.

src/utils/handlebars.ts
hilsonshrestha: Good catch, fixed in 6ec4f3ce. `getAnswerBearingSequence` now matches dynamic iterations by `parseTrialOrder(answer.trialOrder).step === index` instead of a string prefix on the identifier, so a later static component whose name happens to start with the same text (your `block`/`block_1_trial` example) can no longer be swept into the dynamic block's iterations. Added a regression test that reproduces exactly that collision — it fails against the old prefix-matching code and passes with the fix.

src/controllers/ComponentController.tsx
```diff
@@ -280,7 +281,7 @@ export function ComponentController() {
       </Center>
     );
   }
-  const instruction = currentConfig?.instruction || '';
+  const instruction = compileTemplate(currentConfig?.instruction || '', currentConfig?.parameters ?? {});
```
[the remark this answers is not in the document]
hilsonshrestha: Fixed — `instruction` in ComponentController.tsx is now wrapped in `useMemo` keyed on `currentConfig?.instruction`, `currentConfig?.parameters`, and `templateData`, so it only recompiles when one of those actually changes.

src/controllers/MarkdownController.tsx
```diff
@@ -22,9 +31,14 @@ export function MarkdownController({ currentConfig }: { currentConfig: MarkdownC
...
+
+  const renderedText = useMemo(
+    () => compileTemplate(importedText, currentConfig.parameters ?? {}),
+    [importedText, currentConfig.parameters],
+  );
 
   return loading || foundAsset
-    ? <ReactMarkdownWrapper text={importedText} />
+    ? <ReactMarkdownWrapper text={renderedText} />
     : <ResourceNotFound path={currentConfig.path} />;
```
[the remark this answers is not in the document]
hilsonshrestha: Fixed — MarkdownController's 404 message now uses the resolved `templatedPath` instead of the raw `currentConfig.path`.

src/controllers/ImageController.tsx
```diff
@@ -6,14 +6,20 @@ import { ImageComponent } from '../parser/types';
...
+    if (templatedPath.startsWith('http')) {
+      return templatedPath;
     }
-    return `${PREFIX}${currentConfig.path}`;
-  }, [currentConfig.path]);
+    return `${PREFIX}${templatedPath}`;
+  }, [templatedPath]);
 
   const [loading, setLoading] = useState(true);
   const [assetFound, setAssetFound] = useState(false);
```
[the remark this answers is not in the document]
hilsonshrestha: Fixed — ImageController's 404 message now uses the resolved `templatedPath` instead of the raw `currentConfig.path`.

src/utils/handlebars.ts
```diff
@@ -0,0 +1,10 @@
+import Handlebars from 'handlebars';
+
+export function compileTemplate(text: string, parameters: Record<string, unknown> = {}, options?: { noEscape?: boolean }): string {
+  try {
+    return Handlebars.compile(text, { noEscape: options?.noEscape })(parameters);
+  } catch (e) {
+    console.error('Failed to compile handlebars template', e);
```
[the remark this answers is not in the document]
hilsonshrestha: Addressed — added a `describe('compileTemplate', ...)` block in src/utils/tests/handlebars.spec.ts covering variable substitution, missing variables, default parameters, HTML-escaping vs. `noEscape`, malformed-template fallback, and the `REVISIT` data-frame key.

src/utils/handlebars.ts
hilsonshrestha: Addressed — `lookupAnswers`/`lookupAnswersRel` now build an answer-bearing sequence (`getAnswerBearingSequence`) that skips the `end` sentinel and expands a dynamic block's step into its recorded per-iteration answers (ordered by `funcIndex` via `trialOrder`), with `findCurrentPosition` pinning the current iteration via `currentComponent`/`funcIndex` for relative lookups. Added runtime-shaped regression tests (with the `end` sentinel and dynamic-block identifiers) in handlebars.spec.ts.

src/utils/tests/handlebars.spec.ts
hilsonshrestha: Addressed together with the sibling comment on handlebars.ts — added 'lookupAnswers/lookupAnswersRel helper (runtime-shaped sequence)' describe blocks using the real flatSequence shape (with a trailing `end`) plus dynamic-block identifiers and funcIndex-based iteration ordering.

src/components/interface/AppNavBar.tsx
hilsonshrestha: Addressed in 7ce7f26d — AppNavBar now overlays `useStoredAnswer()`'s parameters (`status?.parameters ?? currentConfig?.parameters ?? {}`) the same way ComponentController already did, and HelpModal does the same for help text. Added regression tests in AppNavBar.spec.tsx/HelpModal.spec.tsx covering a dynamic trial's runtime parameters.

src/components/interface/AppNavBar.tsx
hilsonshrestha: Same fix as the sibling comment on this file — 7ce7f26d. AppNavBar (and HelpModal) now reuse `useStoredAnswer()`'s parameters, matching the merge ComponentController already applied to the main surface, with regression tests covering a dynamic trial.

package.json
hilsonshrestha: Fixed in 73c1d1b3 — synced package-lock.json with the Handlebars dependency.

public/demo-html/assets/introduction.md
hilsonshrestha: Fixed — removed the debug lines from the existing HTML demo.

public/demo-templating/assets/quiz-help.md
hilsonshrestha: Fixed — removed the trailing whitespace.

JackWilb (approved):
Reviewed current head e870e3806425bbb8b26476dccfde753f8f803d9e. The empty-image regression is fixed and covered; focused tests, typecheck, lint, lockfile drift, and diff checks pass.

PR #1401, opened 2026-08-20
Add regex validation

jaykim1213:
### Does this PR close any open issues?
Towards #486 

### Give a longer description of what this PR addresses and why it's needed
1. Add following to `ShortTextResponse` and `LongTextResponse`
- Minimum char length
- Maximum char length
- Minimum word length
- Maximum word length
- Regex
- Contains
- Does not contain
- Equals
- Does not equal

```json
{
  "id": "text-validation-regex",
  "type": "shortText",
  "prompt": "Enter ABC-123",
  "textValidation": [
    {
      "type": "matchesRegex",
      "value": "^[A-Z]{3}-\\d{3}$"
    }
  ]
},
{
  "id": "text-validation-contains",
  "type": "longText",
  "prompt": "Enter a response that contains “ReVISit”.",
  "textValidation": [
    {
      "type": "contains",
      "value": "ReVISit"
    }
  ]
},
```

2. Add tests
3. Add form validation page to the `demo-form-elements`

### Are there any additional TODOs before this PR is ready to go?
TODOs:
- [ ] Update relevant documentation
- [ ] ...

PR #1404, opened 2026-08-20
Add built in validation

### LINKED ISSUE #751: Built in validations
A common part of form interfaces is providing end user facing form validations. These are things like "this is not a valid email address" or "the provided number is not in a valid range". Google forms has a simple version of this (basically regex for string inputs and min/max for numerics). It would be useful to do some form of validation for feature parity with other providers. In addition to these adding some type of custom functions might be useful? Simplest might be a simple inline js call?

jaykim1213:
Closes #751 

1. Add built in validation to `shortTextResponse`
- Email
  - Validates the common `local@domain.tld` format.
  - Checks dot-atom local-part formatting and domain labels.
- Phone number
   - 7 digits  ~ 15 digits (excludes + and -)
   - accepts + for country code and - between digits 
- US Phone number
   - 10 digits with 2 hyphens  (`000-000-0000`
- URL
   - Checks prefix (http, https)
   - Checks hostname
- Date
- Time
- Country

```json
 {
    "id": "q-country-dropdown",
    "prompt": "Country dropdown example",
    "secondaryText": "Country names are displayed with emoji flags, and ISO alpha-2 codes are stored.",
    "type": "dropdown",
    "placeholder": "Select a country",
    "options": "countries"
  },
 {
    "id": "built-in-validation-email",
    "type": "shortText",
    "builtInValidation": "email"
  },
{
  "id": "built-in-validation-us-phone-number",
  "type": "shortText",
  "prompt": "Enter a 10-digit US phone number with two hyphens.",
  "secondaryText": "For example: 800-000-0000.",
  "placeholder": "800-000-0000",
  "builtInValidation": "usPhoneNumber"
},
{
  "id": "built-in-validation-url",
  "type": "shortText",
  "prompt": "Enter an HTTP or HTTPS URL.",
  "secondaryText": "For example: https://revisit.dev.",
  "placeholder": "https://revisit.dev",
  "builtInValidation": "url"
},
```
2. Add tests
3. Add built in validation examples to the `demo-form-elements`

TODOs:
- [ ] ...


jaykim1213:
- [x] Add international phone number as well : simple check (min, max length)
- [x] Add country picker
- [x] Add date picker
- [x] Remove postal code, US States

src/components/response/builtInValidation.ts
```diff
@@ -0,0 +1,101 @@
...
+  ['MO', 'Missouri'], ['MT', 'Montana'], ['NE', 'Nebraska'], ['NV', 'Nevada'],
+  ['NH', 'New Hampshire'], ['NJ', 'New Jersey'], ['NM', 'New Mexico'], ['NY', 'New York'],
+  ['NC', 'North Carolina'], ['ND', 'North Dakota'], ['OH', 'Ohio'], ['OK', 'Oklahoma'],
+  ['OR', 'Oregon'], ['PA', 'Pennsylvania'], ['RI', 'Rhode Island'], ['SC', 'South Carolina'],
+  ['SD', 'South Dakota'], ['TN', 'Tennessee'], ['TX', 'Texas'], ['UT', 'Utah'],
+  ['VT', 'Vermont'], ['VA', 'Virginia'], ['WA', 'Washington'], ['WV', 'West Virginia'],
+  ['WI', 'Wisconsin'], ['WY', 'Wyoming'],
+]);
+
+const emailValidation = isEmail();
```
JackWilb: [P2] Please tighten the email validator. Mantine's isEmail() regex accepts malformed dot-atom/domain forms such as .a@example.com, a..b@example.com, a@example..com, and a@-example.com, even though this validator is documented as a valid local@domain.tld check. Add deterministic local-part/domain-label checks and regression cases alongside the existing rejection table.

src/components/response/builtInValidation.ts
```diff
@@ -0,0 +1,101 @@
...
+
+function isHttpUrl(value: string) {
```
JackWilb, line 54: [P2] Please validate the URL prefix and hostname labels, not only new URL() success. The current check accepts http:foo, http://., https://example..com, and hosts beginning or ending with -, despite the documented http:// and https:// format. Require an explicit HTTP(S) // prefix and validate labels while preserving intentional localhost/IP support; add regression tests.

JackWilb (changes requested):
Request changes: the two new built-in validators accept malformed email addresses and malformed HTTP(S) URL inputs. These findings are limited to the PR #1404 delta over b142d4a; inherited PR #1401 behavior is intentionally not included. Please add the validator fixes and regression tests.

src/components/response/builtInValidation.ts
jaykim1213: Tightened the email validation

src/components/response/builtInValidation.ts
jaykim1213: Tightened the URL validation

jaykim1213:
- [x] Update Time picker (12/24 hours, seconds/minutes/hour)
https://mantine.dev/dates/time-picker/#with-dropdown

src/components/response/builtInValidation.ts
```diff
@@ -0,0 +1,104 @@
...
+    return true;
+  }
+
+  const labels = hostname.split('.');
+  return labels.length >= 2 && labels.every((label) => DOMAIN_LABEL_PATTERN.test(label));
+}
+
+function isHttpUrl(value: string) {
+  // Check if the value is a valid HTTP or HTTPS URL
+  if (value.trim() !== value || !/^https?:\/\//i.test(value)) {
```
JackWilb: [P2] Please reject whitespace and control characters anywhere in the raw URL before calling `new URL()`. This check only rejects leading/trailing whitespace, while the URL parser normalizes internal whitespace, so values such as `https://example.com/a b` (and tabs/newlines in the path) pass validation. Add regression cases for spaces, tabs, CR, and LF.

src/utils/dateTimeValidation.ts
```diff
@@ -0,0 +1,30 @@
+function createDate(year: number, month: number, day: number) {
+  if (year < 1) {
+    return null;
+  }
+
+  const date = new Date(0);
```
JackWilb, line 7: [P3] Please make calendar validation independent of the runtime timezone. Constructing and checking this `Date` in local time makes valid input environment-dependent: `parseMonthDayYear("12/30/2011")` returns a date under UTC but `null` under `TZ=Pacific/Apia`, where that local civil day was skipped. Validate the year/month/day with UTC or timezone-neutral arithmetic and add a timezone regression test.

JackWilb (changes requested):
Thanks for tightening the original email and URL cases. Those examples are now covered, but these two smaller edge cases still need fixes before merge.

jaykim1213:
- [x] Add error message when invalid date

src/components/response/builtInValidation.ts
jaykim1213: Added rejecting whitespaces and control characters

src/utils/dateTimeValidation.ts
jaykim1213: Updated calendar validation to be  independent of the runtime timezone

jaykim1213:
- [x] Add `TimePicker` (24 hour format) after #1418  is merged

jaykim1213:
@codex

jaykim1213:
@codex

src/parser/types.ts
```diff
@@ -528,6 +606,8 @@ export interface ShortTextResponse extends BaseResponse {
   maxWordLength?: number;
   /** Validation rules applied to the response value in array order. */
   textValidation?: TextValidationRule[];
+  /** Applies one predefined format from BuiltInValidationType to the response value. */
+  builtInValidation?: BuiltInValidationType;
```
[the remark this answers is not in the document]
jaykim1213: Added parsing check

src/analysis/individualStudy/stats/ResponseVisualization.tsx
```diff
@@ -200,7 +200,7 @@ export function ResponseVisualization({
     }
 
     // Categorical visualization
-    if (response.type === 'radio' || response.type === 'dropdown' || response.type === 'checkbox' || response.type === 'buttons' || response.type === 'ranking-sublist' || response.type === 'ranking-categorical' || response.type === 'ranking-pairwise') {
+    if (response.type === 'date' || response.type === 'time' || response.type === 'radio' || response.type === 'dropdown' || response.type === 'checkbox' || response.type === 'buttons' || response.type === 'ranking-sublist' || response.type === 'ranking-categorical' || response.type === 'ranking-pairwise') {
```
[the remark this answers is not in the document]
jaykim1213: Added chronological sorting

src/components/response/TimeInput.tsx
```diff
@@ -0,0 +1,59 @@
...
+      disabled={disabled}
+      label={prompt.length > 0 && <InputLabel prompt={prompt} required={required} index={index} enumerateQuestions={enumerateQuestions} infoText={infoText} />}
+      description={secondaryText}
+      radius="md"
+      size="md"
+      value={value}
+      format={response.format ?? '24h'}
+      withDropdown
+      withSeconds={response.withSeconds}
+      min={response.min}
```
JackWilb: [P1] Please keep the raw participant value authoritative here. Mantine's min/max handling can clamp an out-of-range pasted value in onChange while leaving the visible segments showing the pasted value (for example, display 08:30 but store 09:00 for a 09:00-18:00 range). That bypasses the intended ReVISit validation error and makes the displayed answer disagree with the stored answer. Remove these picker bounds and rely on getTimeValidationMessage/ReVISit validation; add a controlled paste regression covering the display, stored value, and error.

package.json
```diff
@@ -52,6 +52,7 @@
     "@visdesignlab/upset2-react": "^1.1.0",
     "ajv": "^8.18.0",
     "arquero": "^5.4.0",
+    "countries-list": "^3.4.1",
```
JackWilb, line 55: [P1] Please obtain and record approval for countries-list, or remove it. The repository instructions prohibit adding new libraries without prior approval, and no approval is visible in the PR artifacts or discussion. Resolving the earlier review thread does not document that approval; an existing approved source should be preferred if approval is unavailable.

src/components/response/DateInput.tsx
```diff
@@ -0,0 +1,137 @@
...
+  const options = response.options ?? 'date';
+  const placeholder = response.placeholder ?? getDateValueFormat(options);
+  const {
+    prompt, required, secondaryText, infoText,
+  } = response;
+  const {
+    value, onChange, onBlur, onFocus, readOnly, ...answerProps
+  } = answer;
+  const [showInvalidDateError, setShowInvalidDateError] = useState(false);
+  const parsePickerDate = (inputValue: string) => toPickerDateValue(inputValue, options);
```
JackWilb, line 47: [P2] This conversion still sends a civil date through Mantine's local date handling. With TZ=Pacific/Apia, the stored civil value 12/30/2011 can be formatted by the picker as 12/31/2011, so the participant sees/selects a different date than the value ReVISit stores. Keep the civil string authoritative at the picker boundary and add a component-level Apia regression for initial display and round-trip selection; the UTC parser test does not cover this UI path.

src/parser/parser.ts
```diff
@@ -291,37 +294,63 @@ function verifyTextResponseConstraints(
         });
       }
     });
-
     const textValidation = response.textValidation ?? [];
+    if (response.type === 'shortText' && response.builtInValidation) {
+      const fixedValues = [
```
JackWilb: [P2] This fixedValues check validates requiredValue and each equals independently against the built-in validator, but it never compares them to each other. A config with requiredValue first@example.com and equals second@example.com still parses even though runtime validation requires both values. Treat requiredValue plus non-empty equals rules as one fixed-value set, reject distinct values, and add that exact parser regression.

src/components/response/builtInValidation.ts
```diff
@@ -0,0 +1,105 @@
+import { isEmail } from '@mantine/form';
+import type { BuiltInValidationType } from '../../parser/types';
+
+type BuiltInValidation = {
+  passes: (value: string) => boolean;
+  message: string;
+};
+
+const emailValidation = isEmail();
```
JackWilb: [P2] Please remove the narrower Mantine isEmail() prefilter or make the project regex authoritative. The current prefilter rejects local-part characters that the following project pattern explicitly permits, including o'connor@example.com, user!tag@example.com, and customer/department=shipping@example.com. Add those supported addresses to the positive regression table.

src/components/response/DropdownInput.tsx
```diff
@@ -72,6 +72,7 @@ export function DropdownInput({
         errorProps={{ c: required ? 'red' : 'orange', fz: 'sm', mt: 'xs' }}
         classNames={{ input: classes.fixDisabled }}
         maxDropdownHeight={200}
+        searchable
```
JackWilb: [P3] Please avoid changing keyboard/input behavior for every existing single-select dropdown just to support the large country list. Make searchable conditional on response.options === countries (or document and test global searchability if that broader behavior is intentional).

JackWilb (changes requested):
Request changes: the exact-head re-review confirms the following remaining correctness and scope issues. The earlier malformed URL/email cases, parser timezone validation, date-stat sorting, and stale E2E step are addressed; these items still need resolution.

src/components/response/TimeInput.tsx
jaykim1213: Removed `min` and `max` from the TimePicker to prevent out-of-range values from bypassing ReVISit validation

src/parser/parser.ts
jaykim1213: Added parser validation to reject conflicting `requiredValue` and `equals` value

src/components/response/builtInValidation.ts
jaykim1213: Removed

src/components/response/DropdownInput.tsx
jaykim1213: Updated `searchable` to be optional

src/components/response/DateInput.tsx
jaykim1213: This is an edge case caused by Mantine’s local date handling; fixing it would require replacing`DateInput` with a custom implementation, so I think there is a trade-off

JackWilb (approved):
The country preset value validation fix is pushed in 977d4e31. Parser and runtime validation now reject invalid country codes, with focused coverage; local tests, typecheck, and lint pass.

PR #1406, opened 2026-08-20
Add permanent Firebase and Supabase deployments

### LINKED ISSUE #1231: Deploy to a dedicated supabase instance
revisit.dev/study deploys to our firebase, which makes it easy for us to test whether everything still works. 

We don't have an equal up-to-date repo for supabase, so we should introduce revisit.dev/study-supabase/ 

That should point to the Utah supabase, and be a fork that is updated on releaes, so that the revisit versions are in sync.

JackWilb:
## Summary

- deploy production Firebase and Supabase builds from both `main` and `dev`
- publish the four permanent paths at `/study/`, `/study/dev/`, `/study/supabase/`, and `/study/dev-supabase/`
- use the established one-commit snapshot publisher for main, dev, PR publishing, and PR cleanup
- preserve permanent deployments and PR previews when replacing the `gh-pages` snapshot
- let trusted permanent publishes refresh the shared root `404.html` while preventing PR publishes from changing it
- preserve nested deployment roots when GitHub Pages redirects direct SPA deep links
- build every permanent deployment in Vite production mode so study data uses the existing `prod-` prefix

## Validation

- built all four storage-engine/base-path combinations with `corepack yarn build`
- rebuilt the dev Firebase variant after the single-publisher change and verified its base path, copied `404.html`, and compiled `prod-` prefix
- parsed both workflow YAML files, ran `bash -n` on the publisher, and ran `git diff --check`
- exercised dev, dev-Supabase, PR publish, PR close, main, and main-Supabase transitions against a temporary Git remote
- verified shared `404.html` ownership, stale root deletion, target replacement, sibling preservation, and the final remote snapshot after every transition
- validated root, permanent nested, PR, query/hash, and non-deployment 404 redirect cases

## Deployment note

The Supabase builds assume `https://supabase.revisit.dev` will be repaired. Its current TLS/upstream outage is infrastructure work outside this PR.

Closes #1231

### DIFF
--- a/.github/scripts/publish-pages-snapshot.sh
+++ b/.github/scripts/publish-pages-snapshot.sh
@@ -22,7 +22,11 @@ if [[ ! -d "$pages_checkout/.git" ]]; then
   exit 2
 fi
 
-if [[ -n "$destination_dir" && "$destination_dir" != "dev" && ! "$destination_dir" =~ ^PR[0-9]+$ ]]; then
+if [[ -n "$destination_dir"
+  && "$destination_dir" != "dev"
+  && "$destination_dir" != "supabase"
+  && "$destination_dir" != "dev-supabase"
+  && ! "$destination_dir" =~ ^PR[0-9]+$ ]]; then
   echo "Unsupported Pages destination: $destination_dir" >&2
   exit 2
 fi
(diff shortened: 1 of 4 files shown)


alexsb:
Maybe instead of
```/study/, /study/dev/, /study/supabase/, and /study/dev-supabase/```

We should make the targets
```
/study/
/study-dev/
/study-supabase/
/study-supabase-dev/
```

PR #1412, opened 2026-08-21
Stabilize Playwright replay runs

JackWilb:
## Summary
- Keep Playwright runs tied to the current worktree by starting a fresh Corepack Vite server.
- Default to one worker for deterministic local and CI runs, with environment overrides retained.
- Seek MVNV replay to the recorded end time before asserting restored graph selections.

## Validation
- Chromium Playwright suite: 46/46 passed in 10 consecutive runs (460/460).
- Reported offenders: Cleveland, MVNV, and skip-logic passed 3/3 each.
- Typecheck, lint, diff check, and pre-commit lint passed.

CI status is pending.

### DIFF
--- a/.github/workflows/playwright-webkit-full.yml
+++ b/.github/workflows/playwright-webkit-full.yml
@@ -7,9 +7,13 @@ on:
 
 jobs:
   webkit:
-    name: 🧪 WebKit full E2E suite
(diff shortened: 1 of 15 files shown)


JackWilb:
@codex review

PR #1419, opened 2026-08-21
min max and range validation for several responses as described in #486

### LINKED ISSUE #486: Add response validation for all response types
`responseValidation.ts`
`responseErrors.ts`

# Response Validation Checklist

## General

All response types inherit the applicable General validations unless noted otherwise.

- [x] Required
- [x] Exactly equal (`requiredValue`)
  - Scalar values use exact comparison.
  - Checkbox and multi-dropdown arrays ignore order.
  - Custom response values use deep equality.
  - Matrix and Ranking object values are not currently supported.
- [x] Custom display label for the expected value (`requiredLabel`)
- [x] “I don’t know” validation bypass (`withDontKnow`)

## NumericalResponse

- [x] Minimum value (`min`, greater than or equal)
- [x] Maximum value (`max`, less than or equal)
- [x] Exactly equal (`requiredValue`)
- [x] Greater than (strict)
- [x] Less than (strict)

## ShortTextResponse

- [x] Required
- [x] Exactly equal (`requiredValue`)
- [x] Minimum length
- [x] Maximum length
- [x] Regex
- [x] Includes
- [x] Does not include

## LongTextResponse

- [x] Required
- [x] Exactly equal (`requiredValue`)
- [x] Minimum length
- [x] Maximum length
- [x] Regex
- [x] Includes
- [x] Does not include

## LikertResponse

- [x] Required
- [x] Exactly equal (`requiredValue`)

## DropdownResponse

- [x] Required
- [x] Exactly equal (`requiredValue`)
- [x] Minimum number of selections (`minSelections`)
- [x] Maximum number of selections (`maxSelections`)

## SliderResponse

- [x] Required
- [x] Exactly equal (`requiredValue`)

## RadioResponse

- [x] Required
- [x] Exactly equal (`requiredValue`)
- [x] Require text when “Other” is selected

## CheckboxResponse

- [x] Required
- [x] Exactly equal (`requiredValue`)
- [x] Minimum selection length (`minSelections`)
- [x] Maximum selection length (`maxSelections`)
- [x] Require text when “Other” is selected
- [x] Exact selection length

## MatrixRadioResponse

- [x] Required
- [x] Detect a completely unanswered matrix
- [x] Require every row to be answered
- [x] Detect partially completed matrices

## MatrixCheckboxResponse

- [x] Required
- [x] Detect a completely unanswered matrix
- [x] Require every row to be answered
- [x] Detect partially completed matrices
- [x] Minimum selections per row
- [x] Maximum selections per row

## RankingSublistResponse

- [x] Required
- [x] Detect an empty ranking
- [x] min
- [x] max (rename from numItems)

## RankingCategoricalResponse

- [x] Required
- [x] Detect an empty ranking
- [x] Minimum items per category
- [x] Maximum items per category 
- [x] numItems behavior -> total number of items that should be used
- [x] Require every item to be categorized 

## RankingPairwiseResponse

- [x] Required
- [x] Require at least one complete pair
- [x] Reject unfinished pairs
- [x] Reject invalid pair locations
- [x] Reject unconfigured options
- [x] Reject self-pairs
- [x] Reject duplicate pairs
- [x] Minimum number of pairs
- [x] Maximum number of pairs

## CustomResponse

- [x] Required
- [x] Recursive empty-value detection
- [x] Exactly equal using deep equality
- [x] Module-provided custom `validate()` function
- [x] Custom validation error message
- [x] Module-load error validation

## ReactiveResponse

- [x] Completion handled through stimulus validation

A7700:
Towards #486 

Implements min max validation and exact item validation for Numerical Response, MatrixCheckboxResponse, RankingSublistResponse, RankingCategoricalResponse and RankingPairwiseResponse. Also it moved some of the validation logic to the responses and changes the timing to validate only on next button being clicked.


src/components/response/responseValidation.ts
```diff
@@ -331,10 +439,26 @@ export function validateResponse(
...
 
     if (response.type === 'ranking-sublist' || response.type === 'ranking-categorical' || response.type === 'ranking-pairwise') {
-      if (Object.keys(value).length === 0) {
+      const numItems = Object.keys(value).length;
+      const { min, max } = response;
+      if (numItems === 0) {
         return createValidationResult(response, response.required ? 'unanswered' : 'none');
       }
 
+      if (response.type === 'ranking-sublist') {
```
JackWilb, line 662: Request changes: this sublist path now validates only min/max, while existing configs, types, and schemas still define numItems. Because RankingInput also removed the numItems cap, the bundled top-3 study accepts four or more items; the Chromium E2E fails at tests/demo-ranking-widget.spec.ts:190. Please preserve numItems as a compatibility alias or migrate all configs, schema/docs, and tests to max. Also settle one consistent numItems contract for categorical responses.

src/components/response/responseValidation.ts
```diff
@@ -89,17 +89,39 @@ export function checkCheckboxResponseForValidation(
...
 
-  if (min !== undefined && max !== undefined && (numValue < min || numValue > max)) {
-    return `Please enter a value between ${min} and ${max}`;
+  if (failsStrictMax) {
+    return `Please enter a value less than ${strictMax}.`;
   }
-  if (min !== undefined && numValue < min) {
-    return `Please enter a value of ${min} or greater`;
+
+  if (min && max && (failsMin || failsMax)) {
```
JackWilb: Request changes: using truthiness here breaks the existing min: 0, max: 100 case. The combined branch is skipped and the user sees only the max error instead of the established range message; this directly fails tests/demo-form-elements.spec.ts:77 in Chromium. Use min/max definedness checks and avoid changing the established message unless all consumers are intentionally updated.

src/components/response/responseValidation.ts
```diff
@@ -211,6 +299,26 @@ export function checkMatrixResponse(response: MatrixResponse, value: Record<stri
     return 'Please answer all questions in the matrix to continue.';
   }
 
+  const { min, max } = response;
+  if (min !== undefined || max !== undefined) {
+    const requiredAmountOfQuestionsAnswered = expectedQuestionKeys.every((questionKey) => {
+      const rowValue = value[questionKey].split('|').length;
```
JackWilb: Request changes: matrix withDontKnow answers use the "I don't know" sentinel as the sole row value. Counting split segments here treats that bypass as one selected answer, so withDontKnow: true and min: 2 is rejected even though the existing matrix validation contract accepts the sentinel. Exempt sentinel rows from min/max validation and add a regression test.

src/components/response/responseValidation.ts
```diff
@@ -154,6 +176,70 @@ export function getRankingInstanceIndex(instanceId: string, optionValues: Set<st
...
+  let minMaxError = null;
+  for (const category of ['HIGH', 'MEDIUM', 'LOW'] as const) {
+    const count = Object.values(value ?? {}).filter((cat) => cat === category).length;
+    minMaxError = minMaxValidation(min, max, count, response.type);
+    if (minMaxError) {
+      return minMaxError;
+    }
+  }
+
+  const categorizedItems = Object.values(value ?? {}).filter((cat) => ['HIGH', 'MEDIUM', 'LOW'].includes(cat)).length;
```
JackWilb: Request changes: categorizeAll is checked by counting category-valued entries, not by checking the configured option keys. For options A and B, {A: "HIGH", X: "LOW"} passes the count test while B is missing. Validate the exact configured option set and reject unknown or invalid-category entries.

src/parser/types.ts
```diff
@@ -534,6 +538,10 @@ interface BaseMatrixResponse extends BaseResponse {
   questionOptions: (MatrixQuestionOption | string)[];
   /** The order in which the questions are displayed. Defaults to fixed. */
   questionOrder?: 'fixed' | 'random';
+  /** The minimum amount of answers given per row for the matrix. */
+  min?: number;
```
JackWilb: Request changes: placing min/max on BaseMatrixResponse exposes them to matrix-radio, and the shared validator applies the new per-row count to both matrix types. A schema-valid matrix-radio with min: 2 can never pass because each row accepts exactly one value. Keep these fields on MatrixCheckboxResponse, guard the validator by type, and regenerate both schemas.

JackWilb (changes requested):
Request changes: I reviewed head fa98c8f against dev and confirmed five correctness/regression issues. Two are directly covered by the failed Chromium E2E run: the removed legacy ranking cap and the changed zero-bound numerical error message. The inline comments below identify the smallest safe resolutions.

src/components/response/responseValidation.ts
```diff
@@ -428,10 +565,26 @@ export function validateResponse(
...
 
     if (response.type === 'ranking-sublist' || response.type === 'ranking-categorical' || response.type === 'ranking-pairwise') {
-      if (Object.keys(value).length === 0) {
+      const numItems = Object.keys(value).length;
+      const { min, max } = response;
+      if (numItems === 0) {
         return createValidationResult(response, response.required ? 'unanswered' : 'none');
       }
 
+      if (response.type === 'ranking-sublist') {
```
JackWilb, line 662: P1 — Legacy numItems is still ignored for sublist rankings. RankingInput no longer caps this legacy field, and this validator only reads min/max. Existing configs such as public/demo-form-elements/config.json still use numItems: 2, so a participant can submit more ranked items than the study requested. Please use an effective bound such as response.max ?? response.numItems, with max taking precedence, or fully migrate every config/type/schema/documentation before removing the compatibility behavior.

src/components/response/responseValidation.ts
```diff
@@ -288,7 +401,31 @@ export function checkMatrixResponse(response: MatrixResponse, value: Record<stri
   if (unanswered) {
     return 'Please answer all questions in the matrix to continue.';
   }
-
+  if (response.type === 'matrix-checkbox') {
+    const { min, max } = response;
+    if (min !== undefined || max !== undefined) {
+      const requiredAmountOfQuestionsAnswered = expectedQuestionKeys.every((questionKey) => {
+        const rowValue = value[questionKey];
+        if (isMatrixDontKnowValue(rowValue)) {
```
JackWilb: P2 — This treats any option whose value is "I don't know" as the special bypass, even when the study has not enabled withDontKnow. For example, a matrix-checkbox with answerOptions ["I don't know", "A"], withDontKnow false, and min: 2 incorrectly accepts selecting only the ordinary first option. Please require response.withDontKnow before applying the sentinel bypass.

src/components/response/responseValidation.ts
```diff
@@ -428,10 +565,26 @@ export function validateResponse(
     }
 
     if (response.type === 'ranking-sublist' || response.type === 'ranking-categorical' || response.type === 'ranking-pairwise') {
-      if (Object.keys(value).length === 0) {
+      const numItems = Object.keys(value).length;
```
JackWilb: P2 — Sublist validation counts every raw answer key, including stale or unknown keys that RankingInput filters out of the visible list. A restored value such as {A: "0", X: "1"} can therefore satisfy min: 2 even though the participant sees only one configured item. Please reject unknown keys or derive the count from configured option values before applying the bounds.

src/components/response/responseValidation.ts
```diff
@@ -232,6 +255,94 @@ export function getRankingInstanceIndex(instanceId: string, optionValues: Set<st
...
+    }
+  }
+
+  const categorizedItems = entries.length;
+  if (numItems !== undefined && categorizedItems !== numItems) {
+    return `Please categorize exactly ${numItems} items.`;
+  }
+
+  if (categorizeAll && configuredOptionValues.size > 0) {
+    const missingOptionKeys = [...configuredOptionValues].filter((optionValue) => !(optionValue in (value ?? {})));
```
JackWilb: P2 — Using the `in` operator allows inherited properties to satisfy categorizeAll. With options ["A", "toString"] and value {A: "HIGH"}, the inherited Object.prototype.toString property makes the missing option appear present. Please use an own-key check such as Object.hasOwn or compare against Object.keys(value), and add a reserved-key regression test.

src/parser/types.ts
```diff
@@ -915,6 +923,12 @@ export interface RankingResponse extends BaseResponse {
   default?: Record<string, string>;
   /** The number of items to rank. Applies only to sublist and categorical ranking widgets. */
   numItems?: number;
+  /** The minimum number of items to rank. For sublist ranking it is the number of items, for categorical ranking it is items per category, and for pairwise ranking it is the number of pairs. */
+  min?: number;
+  /** The maximum number of items to rank. For sublist ranking it is the number of items, for categorical ranking it is items per category, and for pairwise ranking it is the number of pairs. */
+  max?: number;
+  /** Whether all items need to be categorized. Defaults to false. */
+  categorizeAll?: boolean;
```
JackWilb, line 1036: P3 — categorizeAll is exposed by the shared RankingResponse type and generated schema for sublist and pairwise responses, but runtime behavior only applies it to categorical rankings. This permits a schema-valid setting that silently does nothing. Please narrow the field to categorical responses or reject it for the other ranking types during parsing.

src/components/response/RankingInput.tsx
```diff
@@ -683,13 +677,16 @@ export function RankingInput({
...
+    max,
+    categorizeAll,
   } = response;
 
   const [localError, setError] = useState<string | null>(null);
   const displayError = localError || error;
 
   const componentProps = {
-    disabled, options, answer, responseId: response.id, numItems,
+    disabled, options, answer, responseId: response.id, numItems, min, max, categorizeAll,
```
JackWilb: P3 — min, max, and categorizeAll are passed through componentProps, but none of the child ranking components consume them; validation reads the response directly on Next. Please remove this dead prop plumbing so the component interface reflects the actual data flow.

JackWilb (changes requested):
Request changes: I re-reviewed the current head and consolidated the remaining issues below. The first four affect runtime correctness or compatibility; the last two are schema/API contract and unnecessary-prop cleanup concerns.

src/components/response/responseValidation.ts
```diff
@@ -232,6 +255,94 @@ export function getRankingInstanceIndex(instanceId: string, optionValues: Set<st
...
+    const count = entries.filter(([, cat]) => cat === category).length;
+    minMaxError = minMaxValidation(min, max, count, response.type);
+    if (minMaxError) {
+      return minMaxError;
+    }
+  }
+
+  const categorizedItems = entries.length;
+  if (numItems !== undefined && categorizedItems !== numItems) {
+    return `Please categorize exactly ${numItems} items.`;
```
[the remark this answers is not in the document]
JackWilb: This is intentional per issue #486: categorical numItems is defined as the exact total number of items used, while min/max apply per category. The validator therefore correctly rejects a response with more than numItems categorized items, even if the UI can place that many items across categories.

src/parser/types.ts
```diff
@@ -908,13 +916,30 @@ export interface CheckboxResponse extends BaseResponse {
  * ```
 */
 export interface RankingResponse extends BaseResponse {
```
JackWilb: This interface is now acting as the shared base for the three concrete ranking variants, but existing consumers still import `RankingResponse` as the discriminated public union. Please rename this shared interface to something like `BaseRankingResponse`, have the three subtype interfaces extend it, and add the exported alias after those declarations: `export type RankingResponse = RankingSublistResponse | RankingCategoricalResponse | RankingPairwiseResponse;`. That preserves the existing API while fixing the TS2322 errors caused by `RankingResponse` no longer having a required `type` field.

JackWilb (changes requested):
Request changes: the public ranking union is still missing on the current head. The inline comment shows the smallest compatible fix.

JackWilb (approved):
Approved: the missing RankingResponse union alias is restored, the shared base is separated, and the affected consumers and generated schemas are updated.

Issue #1387, opened 2026-08-13
Improve Between-subject experiments using a single URL


alexsb:
Two things: 

For a between subjects study, participants should be assigned to one of the conditions so that the assignment is balanced based on how many subjects already completed a condition. 

E.g., if A = 20, B = 10; then the 10 next participants should be assigned to B. 

Here, completed should possibly be something like actually completed PLUS ongoing studies; but not incomplete ones that have been started e.g., 2h ago and are likely to remain incomplete. 

Furthermore, the analytics interface could add a switch where recruiting to a condition could be turned off. 

Context / original comment by @abhsarma see below

------------

In Qualtrics, between-subject studies have a single URL and participants are assigned a condition based on the randomizer. 

The obvious limitation of this approach is that we cannot guarantee an equal number of participants in each condition. However, Qualtrics also provides a “count” dialog box which shows the number of participants who were assigned to each between-subjects conditions. If we want to selectively recruit participants to only a specific condition, we then edit the “count” so that participants do not get assigned to the conditions which already have the desired number of participants.

A better implementation on ReVISit would be in the “Analyze & Manage Study” tab, where we can provide statistics / visualizations of how many participants have completed the study in each condition, as well as provide toggles for each between-subject condition. By default, all the toggles are “on”, but we can selectively turn them off if we want to recruit participants to a specific condition / set of conditions.

abhsarma:
One way to determine incomplete would be Prolific's own time out (which is based on the study duration). If we set the average time to be 20mins, Prolific sets the maximum time to be around 50 mins. So we can maybe add a small multiplier to that, and assume that everyone who started before that and hasn't completed it will not complete it.

PR #1424, opened 2026-08-21
Add stage participant controls and startup preview

### LINKED ISSUE #194: Add field for target number of participants
We should add a field where users can specify how many participants they expect for a complete study. 

This won't be used to turn off participation (that has to be handled by prolific, etc.) but it will be useful to show progress in the debug interface.

ZachCutler04:
## Summary
- add admin timeout controls plus stage capacity and between-subject condition management
- prevent stale participant writes from clearing an admin rejection
- show a safe static first page only after it is proven assignment-independent
- keep preview mode free of recording, device, browser-history, route, and timer side effects

## Validation
- yarn typecheck
- yarn lint (pre-existing StepsPanel.tsx hook-dependency warning only)
- focused storage and static-preview Vitest tests

## Known limitation
Stage capacity is enforced for ordinary entry, but exact simultaneous final-slot admission still needs a backend-atomic reservation transaction/RPC that works for both Firebase and Supabase. This PR does not claim to solve that distributed race.

A broader Node-only edge-case run still has two existing Blob/FileReader environment failures in asset-upload tests.

### DIFF
--- a/src/storage/engines/types.ts
+++ b/src/storage/engines/types.ts
@@ -46,6 +46,7 @@ export type SequenceAssignment = {
   isDynamic: boolean; // Whether the study contains dynamic blocks
   stage: string; // The stage of the participant in the study
   conditions?: string[]; // The study condition(s) assigned to this participant.
+  betweenSubjectsCombinationKey?: string; // The between-subjects combination assigned to this participant.
 };
 
 export type REVISIT_MODE = 'dataCollectionEnabled' | 'developmentModeEnabled' | 'dataSharingEnabled';
@@ -88,6 +89,13 @@ export class StageNoAvailableConditionsError extends Error {
   constructor(public readonly stageName: string) {
     super(`The ${stageName} stage has no enabled between-subjects conditions`);
     this.name = 'StageNoAvailableConditionsError';
+  }
+}
+
+export class StageOnlyDisabledConditionsHaveCapacityError extends Error {
+  constructor(public readonly stageName: string) {
+    super(`The ${stageName} stage only has capacity in disabled between-subjects conditions`);
+    this.name = 'StageOnlyDisabledConditionsHaveCapacityError';
   }
 }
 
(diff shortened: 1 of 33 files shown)


ZachCutler04:
Codewise everything here is good to go, it all works, open to review

UI wise we should take another pass and discuss, it definitely doesnt make a ton of sense at the moment. Wont be huge changes, but definitely some, and idk what the right solution is currently

alexsb:
This is adressing most of what's in #1387, except for the timeout.

src/storage/engines/types.ts
```diff
@@ -991,46 +1277,79 @@ export abstract class StorageEngine {
...
+        'participantData',
+      );
+      if (isParticipantData(existingParticipant)) {
+        return { participant: existingParticipant, modes: null };
+      }
+
+      const { modes, stageData } = await this.getModesAndStageData(this.studyId!);
+      const currentStage = stageData.currentStage.stageName;
+      const currentStageInfo = stageData.allStages.find((stage) => stage.stageName === currentStage);
+      if (modes.dataCollectionEnabled && currentStageInfo?.maxParticipants !== undefined) {
```
JackWilb, line 1304: Could we clean up an assignment if initialization fails after `_getSequence` creates it? This count treats every non-rejected assignment as an occupied slot, but the participant record is saved later (and that write is deferred on cloud engines). A failed write can leave a one-slot stage full, and `rejectParticipant` cannot clear the assignment without a participant record. A rollback or repair path, with a failure test, would keep the slot recoverable.

src/storage/engines/types.ts
```diff
@@ -878,6 +1096,75 @@ export abstract class StorageEngine {
...
+        if (combinationKey) {
+          combinationCounts[combinationKey] = (combinationCounts[combinationKey] || 0) + 1;
+        }
+      });
+
+      const hasRemainingCapacity = (sequence: Sequence) => {
+        const combinationKey = getBetweenSubjectsCombinationKey(sequence.parameters, betweenSubjects);
+        return (combinationCounts[combinationKey] || 0) < desiredParticipantCountsByCombination[combinationKey];
+      };
+      availableSequenceArray = enabledSequenceArray.filter(hasRemainingCapacity);
```
JackWilb, line 1156: Could we choose the eligible condition using its current participant count before choosing a sequence? This filter only removes a condition after it reaches its target. While both are below target, the later `intentIndex` can still assign A when counts are A=20 and B=10; count-aware filtering also does not run without a stage maximum. Issue #1387 asks for new admissions to catch up the less populated condition.

src/utils/getStaticFirstComponent.ts
```diff
@@ -0,0 +1,75 @@
...
+    return null;
+  }
+
+  const componentName = getFirstComponentName(studyConfig.sequence);
+  if (componentName === null) {
+    return null;
+  }
+
+  const component = getComponent(componentName, studyConfig);
+  if ((component?.type !== 'markdown' && component?.type !== 'image')
```
JackWilb, line 69: Could we limit preview to components without responses, or render only the static stimulus? A Markdown or image component can include a custom response. The full response tree mounts `CustomResponseInput` before assignment, and passing `disabled` does not prevent its mount effects. That leaves the preview's side-effect-free guarantee unenforced.

src/storage/engines/types.ts
```diff
@@ -737,18 +883,86 @@ export abstract class StorageEngine {
...
     if (!modesDoc.stage) {
       throw new Error('Stage data not initialized');
     }
 
+    const updatesMaxParticipants = Object.hasOwn(updates, 'maxParticipants');
+    const updatesDisabledBetweenSubjectsCombinations = Object.hasOwn(updates, 'disabledBetweenSubjectsCombinations');
+    const updatesDesiredParticipantsByCombination = Object.hasOwn(updates, 'desiredParticipantsByCombination');
+    const updatesParticipantAssignmentMode = Object.hasOwn(updates, 'participantAssignmentMode');
+    const updatesManualDesiredParticipantsByCombination = Object.hasOwn(updates, 'manualDesiredParticipantsByCombination');
     const updatedAllStages = modesDoc.stage.allStages.map(
```
JackWilb, line 912: Could we make stage updates atomic? Each action reads the modes document and rewrites the full `allStages` array. If two admins change separate condition switches or limits at the same time, the later write can erase the earlier one. A transaction or conditional update would preserve both changes.

JackWilb (changes requested):
Thanks for putting these controls together. I found a few cases that could affect admission and preview behavior, with suggested fixes inline. Could you also address the existing [Copilot comment about the first additional stage color](https://github.com/revisit-studies/study/pull/1424#discussion_r3836400118)? The test currently expects the default color instead of the next unused color. I’d be glad to take another look after these are addressed.

PR #1432, opened 2026-08-22
Add Firebase Cloud Function for screen recording conversion

shresthalucky:
## Summary
- Add a Firebase Cloud Function that converts uploaded screen recordings to WebM server-side for cross-browser playback.
- Preserve Firebase download metadata and make conversion isolated to the finalized object generation.
- Deploy the function from the repository root with an explicit Firebase project.

## Test plan
- [x] Deploy function to a Firebase project and confirm screen recordings uploaded from Chrome/Safari/Firefox convert and play back cross-browser.
- [x] Confirm the functions lockfile has no direct-dependency drift.
- [ ] Verify concurrent/revisited recordings preserve the newest object generation.


functions/src/index.ts
```diff
@@ -0,0 +1,92 @@
...
+import ffmpegInstaller from '@ffmpeg-installer/ffmpeg';
+import ffmpeg from 'fluent-ffmpeg';
+import { parse as hjsonParse } from 'hjson';
+
+const firebaseConfig = hjsonParse(process.env.VITE_FIREBASE_CONFIG ?? '{}');
+const BUCKET: string = firebaseConfig.storageBucket;
+
+admin.initializeApp();
+setGlobalOptions({ maxInstances: 5 });
+ffmpeg.setFfmpegPath(ffmpegInstaller.path);
```
JackWilb, line 17: **[P1] Provision ffprobe separately**\n\nThis configures the bundled ffmpeg executable, but fluent-ffmpeg.ffprobe() still resolves ffprobe through PATH. In a deployment without a system ffprobe, the callback converts the probe error to false at lines 24-27, and every recording then takes the silent skip path at lines 59-61. Bundle a platform-appropriate ffprobe binary and call setFfprobePath, then cover this in a deployment/emulator test.

functions/src/index.ts
```diff
@@ -0,0 +1,92 @@
...
+        ffmpeg(tmpInput)
+          .outputOptions('-c', 'copy', '-f', 'webm')
+          .output(tmpOutput)
+          .on('end', () => resolve())
+          .on('error', (err: Error) => reject(err))
+          .run();
+      });
+
+      logger.info(`Uploading ${filePath}`);
+      await bucket.upload(tmpOutput, {
```
JackWilb, line 75: **[P1] Guard the replacement against stale generations**\n\nThe handler ignores event.data.generation and overwrites the same object path without an ifGenerationMatch precondition. If an older finalization event is delayed or retried after a newer upload replaces that path, this invocation can upload its older conversion over the newer recording. Download the event generation explicitly and use a generation-match precondition; treat a failed precondition as a stale event.

functions/src/index.ts
```diff
@@ -0,0 +1,92 @@
...
+    const tmpInput = path.join(os.tmpdir(), fileName);
+    const tmpOutput = path.join(os.tmpdir(), `${fileName}.tmp`);
+    const bucket = admin.storage().bucket(BUCKET);
+
+    try {
+      logger.info(`Downloading ${filePath}`);
+      await bucket.file(filePath).download({ destination: tmpInput });
+
+      if (!await isWebmCopyCompatible(tmpInput)) {
+        logger.info(`Skipping: codecs not compatible with WebM stream copy: ${filePath}`);
```
JackWilb, line 60: **[P2] Keep the converted-metadata contract consistent**\n\nThe PR test plan says non-convertible files are skipped and still marked converted, but this return leaves the original object and metadata untouched. That also makes a probe failure indistinguishable from an intentional terminal skip. Either persist an explicit terminal status while preserving the download metadata, or update the acceptance contract and add coverage for the chosen behavior.

src/storage/engines/FirebaseStorageEngine.ts
```diff
@@ -577,6 +577,8 @@ export class FirebaseStorageEngine extends CloudStorageEngine {
     participantId: string,
   ): Promise<string | null> {
     const storage = getStorage();
+
+    // Fetches webm converted by firebase function
```
JackWilb, line 581: **[P2] Do not expose the source while conversion is pending**\n\nThe browser upload, asynchronous function output, and analysis read all use the same object path. A reader can therefore fetch the original recording after upload but before the function replaces it, so this path does not reliably return the converted asset. Use distinct source/final object identities or make the read path wait for terminal conversion metadata.

functions/README.md
```diff
@@ -0,0 +1,61 @@
...
+gcloud projects describe YOUR_PROJECT_ID --format="value(projectNumber)"
+```
+
+Replace `YOUR_PROJECT_ID` with your Firebase project ID. Save the output — you'll use it as `PROJECT_NUMBER` in the next steps.
+
+### 2b. Grant Storage Admin to the Eventarc service account
+
+```bash
+gcloud projects add-iam-policy-binding YOUR_PROJECT_ID \
+  --member="serviceAccount:service-PROJECT_NUMBER@gcp-sa-eventarc.iam.gserviceaccount.com" \
```
JackWilb, line 45: **[P2] Document the correct least-privilege IAM principals**\n\nThis grants project-wide roles/storage.admin to the Eventarc service agent, although the function runtime is the identity that performs the Storage download/upload. The direct Cloud Storage Eventarc setup also requires the Cloud Storage service agent publisher role and the trigger/runtime identities appropriate roles. Please document the correct principals with least privilege, and make deployment project selection explicit instead of relying on an ignored .firebaserc and unqualified firebase deploy.

JackWilb (changes requested):
Request changes on the exact current head e2e86cb03556.\n\nI verified the three existing Codex inline findings and agree they are blockers:\n- temporary-file collisions across overlapping invocations;\n- loss of Firebase download-token metadata when replacing the object;\n- H.264/AAC recordings being skipped instead of transcoded.\n\nAdditional verified blockers are included inline for missing ffprobe, stale-generation overwrites, the asynchronous source/final-readiness race, incorrect IAM setup guidance, and the stated converted-metadata contract.\n\nPlease address these cases and add function-specific coverage for probe failure, codec decisions, metadata/token preservation, invocation isolation, and stale-event handling before re-requesting review.

src/storage/engines/FirebaseStorageEngine.ts
JackWilb: One important consequence is caching: the upload path applies a one-year public cache immediately after the source upload. If analysis fetches that URL before conversion finishes, a browser or CDN can keep serving the original bytes even after the function replaces the object. Please keep the source non-cacheable or make the reader wait for the converted state.

functions/package.json
```diff
@@ -0,0 +1,36 @@
...
+    "start": "npm run shell",
+    "deploy": "firebase deploy --only functions",
+    "logs": "firebase functions:log"
+  },
+  "engines": {
+    "node": "22"
+  },
+  "main": "lib/index.js",
+  "dependencies": {
+    "@ffmpeg-installer/ffmpeg": "^1.1.0",
```
JackWilb, line 18: [P1] Use a maintained FFmpeg build

This package resolves the Linux deployment to FFmpeg 4.1.0 from 2018. The function parses participant-uploaded media, so malformed media can reach a very old parser. Please replace it with a maintained, security-patched build and verify that the deployed binary still supports VP9 and Opus.

functions/src/index.ts
```diff
@@ -0,0 +1,88 @@
...
+function isPreconditionFailure(error: unknown): boolean {
+  return typeof error === 'object'
+    && error !== null
+    && 'code' in error
+    && error.code === 412;
+}
+
+export const convertScreenRecording = onObjectFinalized(
+  {
+    bucket: BUCKET, memory: '1GiB', timeoutSeconds: 60, maxInstances: 10,
```
JackWilb, line 29: [P1] Limit video conversion concurrency

At 1 GiB, second-generation functions default to 80 concurrent requests per instance, while this handler starts one CPU-heavy FFmpeg job per request. With maxInstances set to 10, that can overload instances and make the 60-second timeout routine for longer recordings. Set an explicit low concurrency and choose timeout, CPU, and memory from representative recordings.

functions/src/index.ts
```diff
@@ -0,0 +1,88 @@
...
+        },
+      });
+
+      logger.info(`Done: ${filePath}`);
+    } catch (err) {
+      if (isPreconditionFailure(err)) {
+        logger.info(`Skipping stale recording event: ${filePath}`);
+        return;
+      }
+      logger.error(`Conversion failed for ${filePath}`, err);
```
JackWilb, line 80: [P1] Retry recoverable conversion failures

This handler rethrows download, FFmpeg, and upload failures, but the trigger does not enable retries. A transient failure therefore leaves the original object without a converted or failed state. Please enable retries only with idempotent generation checks and record a terminal status for permanent input failures.

functions/.env
```diff
@@ -0,0 +1,10 @@
+VITE_FIREBASE_CONFIG='
```
JackWilb, line 1: [P2] Keep the bucket and deployment project aligned

The committed function environment still hardcodes revisit-utah while the README now tells operators to deploy any project. Deploying with another --project leaves this trigger reading and writing against revisit-utah. Please derive the bucket from the deployed Firebase configuration or require one explicit project setting, and remove the duplicate unqualified deploy path.

functions/package.json
```diff
@@ -0,0 +1,36 @@
...
+  },
+  "devDependencies": {
+    "@types/fluent-ffmpeg": "^2.1.27",
+    "@types/hjson": "^2.4.6",
+    "@typescript-eslint/eslint-plugin": "^8.0.0",
+    "@typescript-eslint/parser": "^8.0.0",
+    "eslint": "^8.9.0",
+    "eslint-config-airbnb-base": "^15.0.0",
+    "eslint-plugin-import": "^2.25.4",
+    "firebase-functions-test": "^3.4.1",
```
JackWilb, line 32: [P2] Add function-specific regression coverage

This package includes firebase-functions-test but has no test script or function tests, and the existing root CI and lockfile workflow do not cover this directory. The generation guard and metadata/conversion behavior need focused tests plus a CI job under Node 22 before we can rely on this function.

functions/src/index.ts
```diff
@@ -0,0 +1,88 @@
...
+import * as fs from 'fs/promises';
+import ffmpegInstaller from '@ffmpeg-installer/ffmpeg';
+import ffmpeg from 'fluent-ffmpeg';
+import { parse as hjsonParse } from 'hjson';
+
+const firebaseConfig = hjsonParse(process.env.VITE_FIREBASE_CONFIG ?? '{}');
+const BUCKET: string = firebaseConfig.storageBucket;
+
+admin.initializeApp();
+setGlobalOptions({ maxInstances: 5 });
```
JackWilb, line 16: [P3] Remove the unused global instance limit

This global maxInstances value is immediately overridden by the function's maxInstances setting, so it has no effect today. Please remove the unused global setting and the unused tsconfig.dev.json if it is not needed, keeping the new function surface minimal.

JackWilb (changes requested):
The final independent review found the remaining issues called out inline on commit 3f0f2b1. The earlier findings about temporary-file isolation, Firebase tokens, Safari transcoding, ffprobe, and stale generations are addressed. I also added the cache consequence to the existing readiness thread. Please address these remaining correctness, deployment, and coverage issues before re-requesting review.

PR #1433, opened 2026-08-22
Cache waveform peaks and use direct URLs for faster replay loads

shresthalucky:
## Summary
- Cache computed waveform peaks in storage and reuse them on replay instead of re-decoding audio every load
- Use direct download URLs for audio/screen recording instead of fetching full blobs
- Stop marking the waveform as loaded for screen-only recordings (no audio track to visualize)

## Test plan
- [x] Open a replay with existing audio; confirm waveform loads faster on second view (peaks cached)
- [x] Open a replay with screen-only recording (no audio); confirm no waveform/mic UI shown and no error


src/components/audioAnalysis/AudioProvenanceVis.tsx
```diff
@@ -303,21 +304,33 @@ export function AudioProvenanceVis({
             throw new Error('Participant ID is required to load audio');
           }
 
-          const [audioUrl, screenUrl] = await Promise.all([
-            safe(storageEngine.getAudio(taskName, participantId)),
-            safe(storageEngine.getScreenRecording(taskName, participantId)),
+          const [audioUrl, cachedPeaks] = await Promise.all([
+            safe(storageEngine.getAudioUrl(taskName, participantId)),
```
JackWilb, line 313: [P1] Update the unit mocks and add focused coverage for this new path. The exact head unit suite still fails because ScreenRecordingReplay and ThinkAloudAnalysis test doubles/assertions provide getScreenRecording while production now calls getScreenRecordingUrl. Please migrate those mocks and cover waveform cache hit/miss, save failure, malformed/missing cache, and screen-only behavior so the new replay logic is exercised.

src/components/screenRecording/ScreenRecordingReplay.tsx
```diff
@@ -45,7 +45,7 @@ export function ScreenRecordingReplay() {
             if (!participantId) {
               throw new Error('Participant ID is required to load audio');
             }
-            const url = await storageEngine.getScreenRecording(identifier, participantId);
+            const url = await storageEngine.getScreenRecordingUrl(identifier, participantId);
```
JackWilb, line 48: [P2] Use screen-recording wording in the missing-participant error. This branch now loads getScreenRecordingUrl, but the adjacent error still says it is loading audio. Please change it to `Participant ID is required to load screen recording` (and keep the failure path from producing an unhandled effect rejection).

src/storage/engines/types.ts
```diff
@@ -1067,6 +1074,16 @@ export abstract class StorageEngine {
     return emptyTags;
   }
 
+  async getWaveformPeaks(taskName: string, participantId: string): Promise<WaveformPeaks | null> {
+    const result = await this._getFromStorage(`audio/${participantId}_${taskName}`, 'waveform.peaks.json');
+    if (!result || !('peaks' in result)) return null;
```
JackWilb: [P2] Validate the complete cached-peaks shape before returning it. Checking only for a `peaks` property accepts malformed arrays or an invalid/missing duration; waveSurfer.load can then reject and the catch path marks valid audio unavailable instead of falling back to uncached decoding. Please treat invalid cache data as a miss and invalidate or version this entry when the source recording is overwritten so stale peaks are not reused.

JackWilb (changes requested):
Thanks for the follow-up. I re-reviewed the current head against origin/dev. Please address the three inline comments below.

I also checked the updated screen-only demo copy from the side chat and am treating that item as addressed; I did not add a duplicate comment for it.

shresthalucky:
@copilot resolve the merge conflicts in this pull request

PR #1434, opened 2026-08-22
Warn when replay browser differs from recording browser

shresthalucky:
## Summary
- Show a warning banner in the replay footer when the browser used to view a screen recording differs from the browser the participant used to record it, since playback may not work properly across browsers

## Test plan
- [x] Record a screen recording in one browser, replay it in a different browser, confirm warning banner appears
- [x] Replay in the same browser as recording, confirm no warning

image


JackWilb (approved):
Verified the replay asset race fix. Audio and screen-recording URLs are keyed by participant/trial, obsolete requests cannot commit state, retained participant metadata is gated, and regression coverage covers late completions and immediate stale-URL exposure.

JackWilb (approved):
Re-approved the updated head. The regression-test mock state is now isolated between tests; the production fix remains keyed and cancellation-safe for participant/trial asset changes.

JackWilb (approved):
Re-approved the cleaned current head. The rebase-only schema URL churn has been removed; the PR diff is limited to the intended ThinkAloud implementation and regression tests.

PR #1440, opened 2026-08-24
Al/1389 deselectable radio

### LINKED ISSUE #1389: Radio buttons not deselectable
**Describe the bug**
If a radio button is selected once it can no longer be deselected but only changed. For required fields this makes sense but if we have a optional quesiton with radiobuttons (because e.g. we only want to allow 1 answer per matrix question) the buttons still can not be deselected which feels weird.

**To Reproduce**
1. Create a matrix component for the study with radio buttons
2. Explicilty set the radiobuttons to `"required": false,`
3. Select one of the radio buttons and try do delete the selection

**Expected behavior**
Radio buttons like check boxes can be deselected (at least when they are optional, as maybe the user does not want to answer or is insecure about his answer after all or it was a premature click).

**Screenshots**
Image

A7700:
Closes #1389 

Currently Radio button selections are not revertable and also for larger input fields all the inputs have to be reverted manually by clicking each element. Now there is a button deselecting all choices for an element and radio buttons can be deselected like in google forms which makes more sense for especially optional radio buttons and clearing all selected choices at once.  
The Button for clear selection onnly appears when at least one selection has been made and disappears once all selections have been cleared or manually deselected. 
The general behavior for checkboxes remains the same, only radio buttons now behave differently.

### DIFF
--- a/src/components/response/RadioInput.tsx
+++ b/src/components/response/RadioInput.tsx
@@ -58,7 +58,18 @@ export function RadioInput({
   return (
     <Radio.Group
       name={`radioInput${response.id}`}
-      label={prompt.length > 0 && <InputLabel prompt={prompt} required={required} index={index} enumerateQuestions={enumerateQuestions} infoText={infoText} />}
+      label={prompt.length > 0 && (
+        <InputLabel
+          prompt={prompt}
+          required={required}
+          index={index}
+          enumerateQuestions={enumerateQuestions}
+          infoText={infoText}
+          clearSelectionButton={(
+            <ClearSelectionButton onClick={() => answer?.onChange?.('')} disabled={disabled} visible={!!answer?.value} />
+          )}
+        />
+      )}
       description={secondaryText}
       key={response.id}
       value={answer?.value}
@@ -169,9 +180,6 @@ export function RadioInput({
           {rightLabel && <Text>{rightLabel}</Text>}
         </Group>
       )}
-      <div style={{ display: 'flex', justifyContent: 'flex-end', marginTop: 8 }}>
-        <ClearSelectionButton onClick={() => answer?.onChange?.('')} disabled={disabled} visible={!!answer?.value} />
-      </div>
     </Radio.Group>
   );
 }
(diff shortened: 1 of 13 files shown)


src/components/response/RadioInput.tsx
```diff
@@ -147,6 +169,9 @@ export function RadioInput({
           {rightLabel && <Text>{rightLabel}</Text>}
         </Group>
       )}
+      <div style={{ display: 'flex', justifyContent: 'flex-end', marginTop: 8 }}>
+        <ClearSelectionButton onClick={() => answer?.onChange?.('')} disabled={disabled} visible={!!answer?.value} />
```
JackWilb: Please preserve the Other text for re-selection in the UI, but exclude it from saved answers unless the main response is selected as `"other"`. This handler only clears the main answer; the separate `${response.id}-other` form field remains in `answerValidator.values`, and `getAnswersFromAllLocations` persists all values. Thus select Other → type text → clear or switch options records stale Other text. Enforce this at the save/export boundary, with regression coverage for both clear and switch-away paths, while retaining the draft text in the form.

JackWilb (changes requested):
Thanks for the clear-selection work. Please address these before merge:

- Preserve the Other text as a local draft so participants do not have to retype it, but omit the `${response.id}-other` field from submitted/persisted answers whenever `${response.id}` is not `"other"` (including clear and switching to another option). The inline comment details the current persistence path.
- Please change the current clear-selection styling to the intended smaller trash icon positioned immediately after the question prompt text, with an accessible label or tooltip and the same disabled/visibility behavior.
- Also fix the one-option `buttons` keyboard case: Mantine wraps arrow navigation and clicks the same card, so the current toggle handler clears a valid answer. Keep deselection through the clear control or guard keyboard navigation.

JackWilb (approved):
Verified the follow-up with a fresh adversarial review. The persistence filter is scoped to configured response IDs and runs only at persistence boundaries; raw validation/check-answer aggregation remains unchanged. Focused tests pass (52/52) and lint passes.

PR #1443, opened 2026-08-24
Add lazy participant auto-timeouts

### LINKED ISSUE #194: Add field for target number of participants

ZachCutler04:
## Summary
- Add lazy, admin-configured participant auto-timeouts.
- Keep timed-out assignments excluded from capacity while retaining late participant data.
- Add distinct timed-out and late-completed analysis statuses.

Closes #194

### DIFF
--- a/src/analysis/individualStudy/LiveMonitor/LiveMonitorView.tsx
+++ b/src/analysis/individualStudy/LiveMonitor/LiveMonitorView.tsx
@@ -1,5 +1,5 @@
 import {
-  Stack, Group, Card, Text, Title, Badge, ActionIcon, Center, Indicator, Tooltip, Button, Flex,
+  Stack, Group, Card, Text, Title, Badge, ActionIcon, Center, Indicator, Tooltip, Button, Flex, Switch, NumberInput, Modal,
 } from '@mantine/core';
 import {
   useMemo, useEffect, useState, useCallback,
@@ -11,12 +11,14 @@ import { StudyConfig } from '../../../parser/types';
 import { StorageEngine, SequenceAssignment } from '../../../storage/engines/types';
 import { ParticipantSection } from './ParticipantSection';
 import { FirebaseStorageEngine } from '../../../storage/engines/FirebaseStorageEngine';
+import { useAuth } from '../../../store/hooks/useAuth';
 
 export interface LiveMonitorParticipantProgress {
   assignment: SequenceAssignment;
   progress: number;
   isCompleted: boolean;
   isRejected: boolean;
+  isTimedOut?: boolean;
 }
 
 export function getFilteredParticipantProgress(
@@ -29,16 +31,20 @@ export function getFilteredParticipantProgress(
       const progress = assignment.total > 0 ? (assignment.answered.length / assignment.total) * 100 : 0;
       const isCompleted = assignment.completed !== null;
       const isRejected = assignment.rejected;
+      const isTimedOut = assignment.autoTimedOutAt !== undefined;
 
       return {
         assignment,
         progress,
         isCompleted,
         isRejected,
+        isTimedOut,
       };
     })
-    .filter(({ isCompleted, isRejected, assignment }) => {
-      const status = isRejected ? 'rejected' : (isCompleted ? 'completed' : 'inProgress');
+    .filter(({
+      isCompleted, isRejected, isTimedOut, assignment,
+    }) => {
+      const status = isRejected ? 'rejected' : (isTimedOut ? 'timedOut' : (isCompleted ? 'completed' : 'inProgress'));
       const statusMatch = includedParticipants.includes(status);
       const stageMatch = selectedStages.includes('ALL') || selectedStages.includes(assignment.stage || '');
 
@@ -48,11 +54,14 @@ export function getFilteredParticipantProgress(
 }
 
 export function groupParticipantProgress(filteredParticipantProgress: LiveMonitorParticipantProgress[]) {
-  const inProgress = filteredParticipantProgress.filter((participant) => !participant.isCompleted && !participant.isRejected);
-  const completed = filteredParticipantProgress.filter((participant) => participant.isCompleted && !participant.isRejected);
+  const inProgress = filteredParticipantProgress.filter((participant) => !participant.isCompleted && !participant.isRejected && !participant.isTimedOut);
+  const completed = filteredParticipantProgress.filter((participant) => participant.isCompleted && !participant.isRejected && !participant.isTimedOut);
   const rejected = filteredParticipantProgress.filter((participant) => participant.isRejected);
+  const timedOut = filteredParticipantProgress.filter((participant) => participant.isTimedOut && !participant.isRejected);
 
-  return { inProgress, completed, rejected };
+  return {
+    inProgress, completed, rejected, timedOut,
+  };
 }
 
 // Progress label components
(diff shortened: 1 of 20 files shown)


src/storage/engines/types.ts
```diff
@@ -1246,6 +1276,35 @@ export abstract class StorageEngine {
...
 
+  private async timeoutExpiredAssignments(
+    modes: ModesAndStageData['modes'],
+  ): Promise<void> {
+    const timeoutMinutes = modes.autoTimeoutMinutes;
+    if (!modes.dataCollectionEnabled || timeoutMinutes === undefined) {
+      return;
+    }
+
+    const deadline = Date.now() - timeoutMinutes * 60_000;
```
JackWilb, line 1287: [P1] Please do not use the participant’s browser clock to decide whether an assignment has expired. This code runs when a new participant opens the study, so a clock that is ahead can time out active participants, while a clock that is behind can leave old assignments consuming capacity. Compare against a trusted backend time or perform the eligibility check in a backend transaction.

src/storage/engines/LocalStorageEngine.ts
```diff
@@ -212,17 +221,37 @@ export class LocalStorageEngine extends StorageEngine {
...
-      await this.studyDatabase.setItem(sequenceAssignmentPath, sequenceAssignments);
-    } else {
-      throw new Error(`Sequence assignment for participant ${participantId} not found`);
-    }
+    await this._runWithLock(`participant-${participantId}`, async () => {
+      const sequenceAssignmentPath = `${this.collectionPrefix}${this.studyId}/sequenceAssignment`;
+      const sequenceAssignments = await this.studyDatabase.getItem<Record<string, SequenceAssignment>>(sequenceAssignmentPath) || {};
+      if (sequenceAssignments[participantId]) {
+        sequenceAssignments[participantId] = {
+          ...sequenceAssignment,
```
JackWilb, line 229: [P1] This spreads the assignment object captured before the lock. If the participant completes, is rejected, or is timed out after that read but before this write, this replaces the newer record and erases that status. Re-read the assignment inside the lock and merge only the `claimed` field. Please add a test for completion racing with slot reuse.

src/storage/engines/SupabaseStorageEngine.ts
```diff
@@ -474,12 +517,72 @@ export class SupabaseStorageEngine extends CloudStorageEngine {
...
-      .update({ data: { ...sequenceAssignment, claimed: true } })
-      .eq('studyId', `${this.collectionPrefix}${this.studyId}`)
-      .eq('docId', sequenceAssignmentPath);
+    await this._runWithLock(`participant-${participantId}`, async () => {
+      // Update the sequence assignment for the participant to mark it as claimed.
+      // The participant lock keeps this whole-record JSON update from racing a
+      // late completion on the source assignment.
+      const { data, error } = await this.supabase
+        .from('revisit')
+        .update({ data: { ...sequenceAssignment, claimed: true } })
```
JackWilb, line 526: [P1] This is the same stale full-record write in the Supabase engine. The lock starts after the caller already selected `sequenceAssignment`, so a late completion, rejection, or timeout can be lost here. Re-read the source row while holding the lock and update only `claimed`, or merge into that fresh row.

src/analysis/individualStudy/table/TableView.tsx
```diff
@@ -92,11 +92,17 @@ export function TableView({
 
           const denominator = Math.max(getSequenceFlatMap(row.sequence).length - 1, 1);
 
-          return { percent: (Object.entries(row.answers).length - incompleteEntries.length) / denominator, completed: row.completed, rejected: row.rejected };
+          return {
+            percent: (Object.entries(row.answers).length - incompleteEntries.length) / denominator,
+            completed: row.completed,
+            rejected: row.rejected,
+            timedOut: !!row.timedOut,
```
JackWilb, line 99: [P2] Auto-timed-out participants are still offered to the existing manual timeout dialog, which filters only incomplete and non-rejected participants. That dialog can then call `rejectParticipant` on this participant, changing the new timed-out status to rejected even though the UI says timed-out participants can finish normally. Please exclude `timedOut` participants from that dialog and add a regression test.

src/analysis/individualStudy/StudyAnalysisTabs.tsx
```diff
@@ -199,11 +208,12 @@ export function StudyAnalysisTabs({ globalConfig }: { globalConfig: GlobalConfig
     if (!expData) return [];
     const expList = Object.values(expData);
 
-    const comp = includedParticipants.includes('completed') ? expList.filter((d) => !d.rejected && d.completed) : [];
-    const prog = includedParticipants.includes('inProgress') ? expList.filter((d) => !d.rejected && !d.completed) : [];
+    const comp = includedParticipants.includes('completed') ? expList.filter((d) => !d.rejected && !d.timedOut && d.completed) : [];
+    const prog = includedParticipants.includes('inProgress') ? expList.filter((d) => !d.rejected && !d.timedOut && !d.completed) : [];
     const rej = includedParticipants.includes('rejected') ? expList.filter((d) => d.rejected) : [];
+    const timedOut = includedParticipants.includes('timedOut') ? expList.filter((d) => !d.rejected && d.timedOut) : [];
```
JackWilb, line 214: [P2] This groups late completions with incomplete timeouts because both have `timedOut === true`. The new `completedLate` field is therefore lost in the main analysis filter, and older consumers such as tidy downloads, snapshot counts, and participant badges still use only the old three statuses. Please define one canonical status mapping and use it everywhere, keeping rejected, timed out, and completed-late distinct.

src/storage/engines/types.ts
```diff
@@ -412,13 +426,25 @@ export abstract class StorageEngine {
...
+  abstract getModes(studyId: string): Promise<RuntimeStudySettings>;
 
   // Sets the mode for the given studyId. The mode is stored as a record with the mode name as the key and a boolean value indicating whether the mode is enabled or not.
   abstract setMode(studyId: string, mode: REVISIT_MODE, value: boolean): Promise<void>;
 
+  async setAutoTimeoutMinutes(studyId: string, minutes: number | undefined): Promise<void> {
+    if (minutes !== undefined && (!Number.isInteger(minutes) || minutes < 1)) {
+      throw new Error('Auto-timeout must be a whole number of minutes greater than zero');
+    }
+    const modes = await this.getModes(studyId);
```
JackWilb, line 438: [P2] This reads and later rewrites the entire settings document to change one field. A concurrent stage or mode update can land between these calls and then be overwritten by this stale copy. Please update only `autoTimeoutMinutes` (including an atomic removal when disabling) with a backend merge or transaction.

src/analysis/individualStudy/LiveMonitor/LiveMonitorView.tsx
```diff
@@ -103,11 +112,54 @@ export function LiveMonitorView({
...
+    storageEngine.getModes(studyId)
+      .then((modes) => {
+        if (!cancelled) {
+          setAutoTimeoutMinutes(modes.autoTimeoutMinutes);
+          setAutoTimeoutDraftMinutes(modes.autoTimeoutMinutes ?? 60);
+          setTimeoutSettingsLoading(false);
+        }
+      })
+      .catch(() => {
+        if (!cancelled) setTimeoutSettingsLoading(false);
```
JackWilb, line 143: [P2] A settings read failure is silently converted into a usable-looking default value. The admin can edit or toggle a value while the backend value is unknown, and a save failure below is also only logged. Please show an error with a retry path and roll back the draft/toggle when a save fails.

src/analysis/individualStudy/summary/utils.tsx
```diff
@@ -56,9 +56,10 @@ function calculateParticipantCounts(visibleParticipants: ParticipantDataWithStat
   const participantCounts: ParticipantCounts = {
     total: filteredParticipants.length,
     // Include !p.rejected to exclude participants rejected manually after completing the study
-    completed: filteredParticipants.filter((p) => p.completed && !p.rejected).length,
-    inProgress: filteredParticipants.filter((p) => !p.completed && !p.rejected).length,
+    completed: filteredParticipants.filter((p) => p.completed && !p.rejected && !p.timedOut).length,
+    inProgress: filteredParticipants.filter((p) => !p.completed && !p.rejected && !p.timedOut).length,
     rejected: filteredParticipants.filter((p) => p.rejected).length,
+    timedOut: filteredParticipants.filter((p) => !p.rejected && p.timedOut).length,
```
JackWilb, line 62: [P2] This intentionally changes the participant-count shape, but the existing `summaryUtils.spec.ts` assertion still expects the old object and fails on this exact head. Please update that expectation and add cases for timed-out and completed-late participants so the new counts are locked down.

src/analysis/individualStudy/management/StageManagementItem.tsx
```diff
@@ -442,6 +445,7 @@ function BetweenSubjectsCombinationTable({
     ...(showParticipantLimits ? [{
       accessorKey: 'desiredParticipants',
       header: 'Total / Maximum',
+      size: 240,
```
JackWilb, line 448: [P3] These table sizing and padding changes are unrelated to auto-timeout behavior. Please remove this styling churn or move it to a separate change so the feature diff stays focused and easier to validate.

JackWilb (changes requested):
This feature has the right intent, but I found several correctness and consistency problems that should be fixed before merging. The existing unresolved comment on the minutes-input blur/switch race is also still valid.

The exact-head checks I ran passed typechecking, but the focused unit run had two failures caused by this change: the summary participant-count shape is stale, and the no-Firebase Live Monitor expectation is stale. A third focused failure is in the unchanged ParticipantTimeoutModal test and is not attributable to this PR. The PR’s `Closes #194` link also does not match the issue: #194 asks for run labels and a target participant count, and explicitly says the target should not turn participation off.

Please address the inline comments, add coverage for the Local/Supabase interleavings and the full status model, and rerun the relevant unit and browser checks before merging.

PR #1449, opened 2026-08-31
auto-advance, delay and block answer change

### LINKED ISSUE #1447: Auto-advance on submitting a response in forms
Responses created in the `config.json` such as buttons should have the option to auto-advance the participant to the next trial. This is necessary for perceptual studies such as the correlation task or the stroop test which require participants quickest response.

Another related option could be to prevent participants from changing their response after they've selected an answer. But this could be potentially unnecessary

For the auto-advance feature, two additional fields under response: 

```
{
  ...
  "autoAdvanceToNextStep": true, // 'false' by default, could be set to 'true'
  "autoAdvanceDelay": 300"
}
```

For disable:

```
{
  ...
  "allowResponseChange": true, // 'true' by default, could be set to 'false'
}
```

A7700:
Closes #1447 

This PR adds the options autoAdvanceToNextStep, autoAdvanceDelay and allowResponseChange to the ButtonsResponse. Studies sometimes want the user to automatically navigate to the next page/question if some input was given especially when a button is pressed. The options introduced allow to automatically proceed to the next page. This procedure can be delayed and the given answer can be locked so it cannot be changed even if the delay is active and the next question is not shown immediately.

### DIFF
--- a/src/components/response/ResponseBlock.tsx
+++ b/src/components/response/ResponseBlock.tsx
@@ -677,23 +677,68 @@ export function ResponseBlock({
     };
   }, [identifier]);
 
+  const userInteractedRef = useRef(false);
+
   useEffect(() => {
-    if (!autoAdvanceOwner || autoAdvanceIdentifierRef.current === identifier) {
-      return;
+    userInteractedRef.current = false;
+  }, [identifier]);
+
+  const handleNextClickRef = useRef(handleNextClick);
+  useEffect(() => {
+    handleNextClickRef.current = handleNextClick;
+  }, [handleNextClick]);
+
+  const prevValuesRef = useRef(answerValidator.values);
+  useEffect(() => {
+    if (!isEqual(prevValuesRef.current, answerValidator.values)) {
+      if (Object.keys(answerValidator.values).length > 0) {
+        userInteractedRef.current = true;
+      }
+      prevValuesRef.current = answerValidator.values;
+    }
+  }, [answerValidator.values]);
+
+  useEffect(() => {
+    if (!autoAdvanceOwner) {
+      return undefined;
     }
 
     const readyResponse = autoAdvanceResponses.find((response) => hasAnswerValue(combinedValues[response.id]));
+
     if (!readyResponse) {
-      return;
+      autoAdvanceIdentifierRef.current = null;
+      return undefined;
+    }
+
+    const hasFeedback = 'provideFeedback' in readyResponse && Boolean(readyResponse.provideFeedback);
+    if (hasFeedback) {
+      return undefined;
+    }
+
+    if (!userInteractedRef.current) {
+      return undefined;
+    }
+
+    if (autoAdvanceTimeoutRef.current) {
+      clearTimeout(autoAdvanceTimeoutRef.current);
+      autoAdvanceTimeoutRef.current = undefined;
     }
 
     autoAdvanceIdentifierRef.current = identifier;
     if (readyResponse.type === 'buttons') {
+      const delay = readyResponse.autoAdvanceDelay ?? 0;
       autoAdvanceTimeoutRef.current = setTimeout(() => {
-        handleNextClick();
-      }, readyResponse.autoAdvanceDelay ?? 0);
+        handleNextClickRef.current();
+      }, delay);
     }
-  }, [autoAdvanceOwner, autoAdvanceResponses, combinedValues, customResponseLoadErrors, customResponseValidators, handleNextClick, identifier]);
+
+    return () => {
+      if (autoAdvanceTimeoutRef.current) {
+        clearTimeout(autoAdvanceTimeoutRef.current);
+        autoAdvanceTimeoutRef.current = undefined;
+      }
+    };
+  }, [autoAdvanceOwner, autoAdvanceResponses, combinedValues, identifier]);
 
   let index = 0;
   return (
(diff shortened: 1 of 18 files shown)


JackWilb:
- Demo form elements
- Remove next by default on auto advance
- Add flag for removing the next button

src/components/response/ResponseBlock.tsx
```diff
@@ -660,6 +660,88 @@ export function ResponseBlock({
...
+    }
+
+    const readyResponse = autoAdvanceResponses.find((response) => hasAnswerValue(combinedValues[response.id]));
+
+    if (!readyResponse) {
+      autoAdvanceIdentifierRef.current = null;
+      return undefined;
+    }
+
+    const hasFeedback = 'provideFeedback' in readyResponse && Boolean(readyResponse.provideFeedback);
```
JackWilb, line 715: This check looks for `provideFeedback` on the response, but training feedback is configured on the component. For a valid training component this is always false, so the timer can advance without Check Answer and skip incorrect-answer attempts. Please gate this path with the same training state used by the Next button.

src/components/response/ResponseBlock.tsx
```diff
@@ -660,6 +660,88 @@ export function ResponseBlock({
...
+  useEffect(() => {
+    userInteractedRef.current = false;
+  }, [identifier]);
+
+  const handleNextClickRef = useRef(handleNextClick);
+  useEffect(() => {
+    handleNextClickRef.current = handleNextClick;
+  }, [handleNextClick]);
+
+  const prevCombinedValuesRef = useRef(combinedValues);
```
JackWilb, line 691: Comparing the shared Redux values treats initialization as a participant interaction. A default or restored answer can therefore arm this timer without a new button selection, including when the participant returns with Previous. Please pass an explicit selection event from the buttons response and arm the timer only for that event.

src/components/response/ResponseSwitcher.tsx
```diff
@@ -76,30 +76,99 @@ export function ResponseSwitcher({
...
+  const responseChangeLocked = useMemo(() => {
+    if (!('allowResponseChange' in response)) {
+      return false;
+    }
+
+    if (response.type === 'buttons' && response.allowResponseChange !== false) {
+      return false;
+    }
+
+    if (response.allowResponseChange === false) {
```
JackWilb, line 160: Clearing a default sends an empty value through the same interaction path, sets the interaction flag, and then locks every option when `allowResponseChange` is false. That leaves a required response empty and uneditable. Only lock after a non-empty option is selected, or make Clear Selection unavailable in this mode.

src/parser/parser.ts
```diff
@@ -773,6 +773,30 @@ function verifyStudyConfig(studyConfig: StudyConfig, importedLibrariesData: Reco
...
 
+  Object.entries(studyConfig.components).forEach(([componentName, component]) => {
+    const resolvedComponent = studyComponentToIndividualComponent(component, studyConfig);
+
+    const responses = resolvedComponent.response ?? [];
+    const hasAutoAdvanceButton = responses.some(
+      (response) => response.type === 'buttons' && response.autoAdvanceToNextStep === true,
+    );
+    const isAutoAdvance = hasAutoAdvanceButton || resolvedComponent.nextButtonAutoAdvanceTime !== undefined;
+    resolvedComponent.nextButtonHidden = resolvedComponent.nextButtonHidden ?? (isAutoAdvance ? true : ((studyConfig.uiConfig as { nextButtonHidden?: boolean })?.nextButtonHidden ?? false));
```
JackWilb, line 784: For an inherited component, `studyComponentToIndividualComponent` returns a merged clone, so this `nextButtonHidden` default is discarded. Runtime resolution then falls back to showing Next, contrary to the auto-advance default. Compute the default at the runtime owner or persist it on the actual component entry, and cover inherited/library cases.

src/components/response/ResponseBlock.tsx
```diff
@@ -660,6 +660,88 @@ export function ResponseBlock({
...
+
+    if (autoAdvanceTimeoutRef.current) {
+      clearTimeout(autoAdvanceTimeoutRef.current);
+      autoAdvanceTimeoutRef.current = undefined;
+    }
+
+    autoAdvanceIdentifierRef.current = identifier;
+    if (readyResponse.type === 'buttons') {
+      const delay = readyResponse.autoAdvanceDelay ?? 0;
+      autoAdvanceTimeoutRef.current = setTimeout(() => {
```
JackWilb, line 732: This direct call bypasses `NextButton`'s `nextButtonEnableTime` and `nextButtonDisableTime` checks. A response can advance before Next is enabled or after it has timed out. Please share the navigation eligibility check with response auto-advance, or reject/document incompatible timer combinations.

src/store/hooks/tests/useNextStep.spec.tsx
```diff
@@ -390,4 +522,63 @@ describe('useNextStep', () => {
     }));
     expect(mockNavigate).toHaveBeenCalledWith('/study-1/1');
   });
+
+  test('locks response options immediately during autoAdvanceDelay before navigation occurs', async () => {
```
JackWilb, line 526: These tests do not mount `ResponseBlock` or `ResponseSwitcher`; they manually invoke `goToNextStep` and create unrelated timers. They would pass if the new feature were removed. Please add focused tests that render the response UI and exercise real selection, delay, training, Previous/default, cross-location, and clear-selection behavior.

JackWilb (changes requested):
Request changes. The main flow needs a few correctness fixes before merge. The inline comments call out the issues affecting training feedback, restored/default answers, response locking, inherited components, navigation timers, and test coverage. Please address those cases and add focused component-level tests for the real user interactions.

PR #1450, opened 2026-09-01
keyboard input controls supported with keymapper

### LINKED ISSUE #1448: Support keyboard inputs for buttons
In psychometric studies (e.g., correlation, stroop test etc.), there are on-screen buttons for user input, but also keyboard-based inputs (e.g., arrows for correlation, numbers for stroop) as these allow users to respond more quickly then maneuvering the mouse.

"response": [
        {
          "id": "response",
          "prompt": "Selected ink color",
          "type": "buttons",
          "options": [
            {"label": "RED", "key": 1},
            {"label": "ORANGE", "key": 2},
            {"label": "YELLOW "key": 3},
            {"label": "GREEN", "key": 4},
            {"label": "BLUE", "key": 5},
            ...
          ],
          "location": "belowStimulus",
          "required": true
        }

A few other considerations:
- For arrows, the "key" should be either "ArrowLeft", "ArrowUp", "ArrowRight", "ArrowDown"?
- For letters, we should make sure that it is case agnostic (so both "a" and "A" should work).
- I am not sure if we should restrict the type of keys that can be used (in other words, restrict it to alphabets, numbers, arrows and space)?

A7700:
Closes #1448 

Custom keyboard controls are not easily accessible in the config files but are really usefull especially when buttons are involved in the questions. Thie PR enables the user to define a mapping or a list in the json to link buttons to keys on the keyboard and on keypress the linked button is selected.

### DIFF
--- a/src/components/NextButton.tsx
+++ b/src/components/NextButton.tsx
@@ -1,4 +1,6 @@
-import { Alert, Button, Group } from '@mantine/core';
+import {
+  Alert, Button, Group, Kbd,
+} from '@mantine/core';
 import {
   JSX, useEffect, useMemo, useRef, useState,
 } from 'react';
@@ -148,6 +150,7 @@ export function NextButton({
           disabled={nextButtonDisabled}
           onClick={() => onNext()}
           px={location === 'sidebar' && checkAnswer ? 8 : undefined}
+          rightSection={nextOnEnter ? <Kbd size="xs">↵ Enter</Kbd> : undefined}
         >
           {label}
         </Button>
--- a/src/components/NextButton.tsx
+++ b/src/components/NextButton.tsx
@@ -110,7 +110,7 @@ export function NextButton({
 
   useEffect(() => {
     const handleKeyDown = (event: KeyboardEvent) => {
-      if (event.key !== 'Enter' || shouldIgnoreEnter(event.target)) {
+      if (event.key !== 'Enter' || event.defaultPrevented || (event as unknown as { __keyMapperHandled?: boolean }).__keyMapperHandled || shouldIgnoreEnter(event.target)) {
         return;
       }
 
@@ -150,7 +150,35 @@ export function NextButton({
           disabled={nextButtonDisabled}
           onClick={() => onNext()}
           px={location === 'sidebar' && checkAnswer ? 8 : undefined}
-          rightSection={nextOnEnter ? <Kbd size="xs">↵ Enter</Kbd> : undefined}
+          aria-label={label}
+          rightSection={nextOnEnter && !onCheckAnswer ? (
+            <Kbd
+              size="xs"
+              aria-hidden="true"
+              style={{
+                backgroundColor: 'transparent',
+                color: 'inherit',
+                boxShadow: 'none',
+                border: 'none',
+                fontSize: '11px',
+                fontWeight: 600,
+              }}
+            >
+              ↵
+            </Kbd>
+          ) : undefined}
+          styles={{
+            inner: { alignItems: 'stretch' },
+            section: nextOnEnter && !onCheckAnswer ? {
+              alignItems: 'center',
+              justifyContent: 'center',
+              padding: '0 10px',
+              marginRight: -16,
+              marginBlock: -1,
+              borderLeft: '1px solid rgba(255, 255, 255, 0.25)',
+              backgroundColor: 'rgba(0, 0, 0, 0.08)',
+            } : undefined,
+          }}
         >
           {label}
         </Button>
--- a/src/components/response/ButtonsInput.tsx
+++ b/src/components/response/ButtonsInput.tsx
@@ -9,6 +9,7 @@ import { useStoredAnswer } from '../../store/hooks/useStoredAnswer';
 import { InputLabel } from './InputLabel';
 import { OptionLabel } from './OptionLabel';
 import { parseStringOptions } from '../../utils/stringOptions';
+import { KeyMapper } from './KeyMapper';
 
 export function ButtonsInput({
   response,
@@ -65,6 +66,12 @@ export function ButtonsInput({
         errorProps={{ c: required ? 'red' : 'orange', fz: 'sm', mt: 'xs' }}
         style={{ '--input-description-size': 'calc(var(--mantine-font-size-md) - calc(0.125rem * var(--mantine-scale)))' }}
       >
+        <KeyMapper
+          options={orderedOptions}
+          keys={response.keyMapping}
+          onSelect={(val) => answer?.onChange?.(val)}
+          disabled={disabled}
+        />
         <Flex justify="space-between" align="center" gap="xl" mt="xs">
           {orderedOptions.map((radio, idx) => (
             <Radio.Card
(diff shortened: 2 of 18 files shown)


JackWilb:
Inline key mapping on the options
Use a kbd element on the button (allow hiding with a config prop)
Keystroke vs click logging in track. These need to be logged as keystrokes in trrack (maybe new issue?)
Parser error for invalid key maps (Any combo, with + between, verify that they're valid keys) — need to choose what the valid keys are called, is there some standard? (https://revisit-nsf.slack.com/archives/C03NLALH0JJ/p1788445984338849)

src/parser/parser.ts
```diff
@@ -617,6 +617,42 @@ function hasConditionalBlockInsideRestrictedOrderAncestor(
...
 
+const ALLOWED_SPECIAL_KEYS = new Set([
+  'space', 'spacebar', 'enter', 'tab', 'escape', 'backspace', 'delete',
+  'arrowleft', 'arrowright', 'arrowup', 'arrowdown', 'home', 'end', 'pageup', 'pagedown',
+]);
+
+function verifyKeyMappings(
+  basePath: string,
+  component: Partial<IndividualComponent>,
+  warnings: ParsedConfig<StudyConfig>['warnings'],
```
JackWilb: This currently records invalid mappings as warnings, so a config containing "14", "RightKeyboardArrow", or a whitespace-only key still loads. The requirement was for a parser error that stops the study. Could this append to errors instead, and validate the agreed grammar there, including empty values, multi-digit numbers, unknown names, and duplicate bindings?

src/components/response/KeyMapper.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+      if ((event as unknown as { __keyMapperHandled?: boolean }).__keyMapperHandled) {
+        return;
+      }
+
+      if (event.ctrlKey || event.metaKey || event.altKey) {
+        return;
+      }
+
+      const pressedKey = (event.key || '').toLowerCase();
```
JackWilb: The parser and runtime do not yet share a representation for combinations such as Shift+X. This code compares only event.key and ignores modifier state, so a plain "x" mapping also fires for Shift+X while "Shift+X" itself cannot match. Please parse and match one canonical modifier-plus-key format.

src/components/response/ButtonsInput.tsx
```diff
@@ -65,19 +78,35 @@ export function ButtonsInput({
         errorProps={{ c: required ? 'red' : 'orange', fz: 'sm', mt: 'xs' }}
         style={{ '--input-description-size': 'calc(var(--mantine-font-size-md) - calc(0.125rem * var(--mantine-scale)))' }}
       >
+        <KeyMapper
```
JackWilb, line 131: KeyMapper is mounted as a sibling of the Radio.Card controls. Its focus guard treats those visible controls as external buttons, so after a participant clicks or tabs to an option, mapped shortcuts are ignored. Please give the mapper explicit ownership of the actual response controls (or bind at the response level) and add a focused-option browser test.

src/components/response/KeyMapper.tsx
```diff
@@ -0,0 +1,143 @@
...
+        return keyLower === pressedKey;
+      };
+
+      // Find the first option whose inline `key` matches the physical key press
+      for (const option of options) {
+        if (typeof option === 'object' && option !== null && option.key) {
+          if (isKeyMatch(option.key)) {
+            (event as unknown as { __keyMapperHandled?: boolean }).__keyMapperHandled = true;
+
+            if (typeof event.preventDefault === 'function') {
```
JackWilb, line 84: A mapped Enter calls preventDefault(), but that does not stop NextButton's separate window listener. The same keydown can therefore select an option and immediately run Check Answer or Next. Please give one handler ownership of Enter (for example, have NextButton ignore defaultPrevented) and cover this with an integration test.

src/components/NextButton.tsx
```diff
@@ -148,6 +150,7 @@ export function NextButton({
           disabled={nextButtonDisabled}
           onClick={() => onNext()}
           px={location === 'sidebar' && checkAnswer ? 8 : undefined}
+          rightSection={nextOnEnter ? <Kbd size="xs">↵ Enter</Kbd> : undefined}
```
JackWilb: This visible Kbd text becomes part of the button's accessible name, changing "Next" to something like "Next ↵ Enter"; current Chromium tests that look for the exact Next button are failing. Please keep the hint visual without changing the accessible name, and make sure the hint is attached to the action Enter actually invokes when Check Answer is active.

src/components/response/KeyMapper.tsx
```diff
@@ -0,0 +1,143 @@
...
+    <div
+      ref={containerRef}
+      tabIndex={-1}
+      style={{ display: 'block', outline: 'none' }}
+    >
+      {/* Screen-reader accessible hidden buttons for keyboard options */}
+      {options
+        ?.filter((option) => option.key)
+        .map((option) => (
+          <button
```
JackWilb: These clipped buttons are still ordinary focusable controls inside the FocusTrap. They duplicate the visible radio options in the tab order and accessibility tree, and the later autofocus moves focus to an empty div. Please remove the duplicate controls/autofocus and expose the shortcut on the real response controls instead.

src/components/response/ButtonsInput.tsx
```diff
@@ -65,19 +78,35 @@ export function ButtonsInput({
...
+          options={orderedOptions}
+          onSelect={(val) => answer?.onChange?.(val)}
```
JackWilb: Keyboard selection and clicking both call the same answer.onChange callback here. ResponseBlock then records both through the generic "Update form field" Trrack action, so the provenance graph cannot tell which interaction caused the answer. Please carry the interaction source through the update and add a provenance regression.

JackWilb (changes requested):
Thanks for adding the inline key mapping and the key hints. I rechecked the latest commits and am requesting changes because the current implementation still has a few correctness and accessibility problems:

- Invalid mappings are warnings, so configs such as "14" still run instead of failing in the parser.
- Shift+X-style combinations are not implemented consistently between validation and runtime.
- Shortcuts stop working after a participant focuses a visible option, and multiple visible responses use first-listener-wins ownership.
- A mapped Enter can also trigger the Next/Check Answer handler.
- The new Next hint changes the accessible name and is causing current Chromium failures.
- Keyboard and click changes are still indistinguishable in Trrack.

The inline comments below describe the smallest fixes I think are needed. Please also add integration coverage for focus ownership, mapped Enter, parser failures/combinations, accessible names, and Trrack interaction source.

A7700:
do not use mantine key elements but instead use some small subtle symbols like copy and collapse symbols on mantine Kbd documentation
image

JackWilb (approved):
Thanks for the keyboard input work. I merged the current `dev` branch and addressed the review follow-ups:

- Mapped button selections now record whether they came from a keyboard shortcut or a click in provenance.
- Config validation rejects Tab shortcuts and Enter shortcuts that conflict with `nextOnEnter`. Unsupported modifier combinations produce warnings.
- Focused action buttons retain their normal Enter and Space activation, and shortcut mappings are limited to button options.
- Removed the unused focus wrapper and kept the updated Stroop test assertions from both branches.

The focused unit tests and Chromium Stroop test passed locally. The pushed commit also passed lint, unit tests, deployment, and all four Chromium CI shards. Approved.

PR #1454, opened 2026-09-03
delay for form elements; fix enumeration with divider elements

### LINKED ISSUE #613: Add delay to UI widgets
I want to specify a delay until an arbitrary form element activates. Specified in milliseconds. 

Most immediately useful for #610

A7700:
Closes #613 

Arbitrary Form elements can now be disabled for a specified amount of time in ms. This PR also fixes the enumeration issue with dividers.

### DIFF
--- a/public/demo-form-elements/config.json
+++ b/public/demo-form-elements/config.json
@@ -48,6 +48,42 @@
           "prompt": "## Numerical Inputs",
           "restartEnumeration": true
         },
+                {
+          "id": "q-numerical-delayed",
+          "prompt": "Delayed numerical example",
+          "secondaryText": "A numeric example that is required to complete and has an option to say 'Don't know'. It will be enabled after a 5 second delay.",
+          "infoText": "This numerical input accepts values between 0 and 100.",
+          "type": "numerical",
+          "placeholder": "Enter a number from 0 to 100",
+          "max": 100,
+          "min": 0,
+          "withDontKnow": true,
+          "delay": 5000
+        },
+        {
+          "id": "q-slider-delayed",
+          "prompt": "Delayed slider example",
+          "secondaryText": "A slider for 0-100 with 3 breakpoints for bad, mid, and good. It will be enabled after a 5 second delay.",
+          "infoText": "Use the slider to indicate your rating on a scale from 'Bad' to 'Good'.",
+          "type": "slider",
+          "startingValue": 50,
+          "spacing": 50,
+          "options": [
+            {
+              "label": "Bad",
+              "value": 0
+            },
+            {
+              "label": "Mid",
+              "value": 50
+            },
+            {
+              "label": "Good",
+              "value": 100
+            }
+          ],
+          "delay": 5000
+        },
         {
           "id": "q-numerical",
           "prompt": "Numerical example",
--- a/src/components/response/ResponseBlock.tsx
+++ b/src/components/response/ResponseBlock.tsx
@@ -46,6 +46,7 @@ import { useManagedTrrack } from '../../store/hooks/useRevisitTrrack';
 import { useStorageEngine } from '../../storage/storageEngineHooks';
 import { showNotification } from '../../utils/notifications';
 import { getAnswersFromAllLocations, getPersistedAnswersFromAllLocations } from '../../utils/getAnswersFromAllLocations';
+import { DelayedResponseWrapper } from './DelayedResponseWrapper';
 
 type Props = {
   status?: StoredAnswer;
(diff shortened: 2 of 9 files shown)


src/components/response/ResponseBlock.tsx
```diff
@@ -672,57 +673,79 @@ export function ResponseBlock({
...
-                      index={index}
-                      config={config}
+                    <DelayedResponseWrapper
+                      delay={response.delay}
                       disabled={disabledAttempts}
-                      errors={errors}
-                    />
+                    >
+                      {(isDelayedDisabled: boolean) => (
+                        <div
```
JackWilb, line 699: This extra div is now inserted around every response, including responses without a delay. The demo-style layout and existing study styles rely on `.responseBlock > div > .response` and on the response being the direct child of the outer wrapper, so these selectors stop matching. Please keep the existing DOM structure and apply the delay state to the existing `.response` element, or otherwise avoid adding a wrapper node.

src/components/response/ResponseBlock.tsx
```diff
@@ -672,57 +673,79 @@ export function ResponseBlock({
...
-                          customResponseLoadErrors[response.id],
-                          { showRequiredErrors: errors },
-                        )
-                        : undefined}
-                      response={response}
-                      index={index}
-                      config={config}
+                    <DelayedResponseWrapper
+                      delay={response.delay}
                       disabled={disabledAttempts}
```
JackWilb, line 696: Passing `disabledAttempts` into this wrapper makes the whole response inert, even in the existing flows where `ResponseSwitcher` intentionally keeps inputs editable before a dynamic block or when the next page has a Previous button. Please keep the timer-disabled state separate from the existing training-state behavior and only make the response inert while its delay timer is active.

public/demo-form-elements/config.json
```diff
@@ -47,6 +47,42 @@
...
+          "id": "q-numerical-delayed",
+          "prompt": "Delayed numerical example",
+          "secondaryText": "A numeric example that is required to complete and has an option to say 'Don't know'. It will be enabled after a 5 second delay.",
+          "infoText": "This numerical input accepts values between 0 and 100.",
+          "type": "numerical",
+          "placeholder": "Enter a number from 0 to 100",
+          "max": 100,
+          "min": 0,
+          "withDontKnow": true,
+          "delay": 5000
```
JackWilb, line 61: This demo still points at the v2.4.3 schema, but that schema does not define `delay` and rejects this new property as an additional property. Please point the example at the current/dev schema so the example validates and editor tooling recognizes the field.

src/parser/types.ts
```diff
@@ -410,6 +410,8 @@ export interface BaseResponse {
   style?: Styles;
   /** Exclude response from randomization. If present, will override the `responseOrder` randomization setting in the components. Defaults to false. */
   excludeFromRandomization?: boolean;
+  /** Delay in milliseconds before an element activates. Defaults to 0 meaning the element is always visible. */
+  delay?: number;
```
JackWilb, line 414: The schema currently accepts negative delays, but the implementation silently treats them as zero. A negative duration has no meaningful interpretation here. Please constrain `delay` to non-negative values in the type metadata and regenerate both schemas.

src/components/response/tests/ResponseBlock.spec.tsx
```diff
@@ -446,6 +446,74 @@ describe('ResponseBlock', () => {
...
+    const wrapperWithStyles = Array.from(questionBlock!.querySelectorAll('div')).find(
+      (el) => el.style.pointerEvents !== '',
+    );
+
+    expect(wrapperWithStyles).toBeDefined();
+    expect(wrapperWithStyles?.style.pointerEvents).toBe('none');
+    expect(wrapperWithStyles?.style.opacity).toBe('0.4');
+  });
+
+  test('does not apply delay styles to non-delayed responses and maintains sequential index', async () => {
```
JackWilb, line 480: The test name says it checks sequential enumeration, but it never asserts either response index; it only checks styles on q1. The divider regression would still pass if numbering were wrong. Please capture the index or render the real labels and cover divider plus restartEnumeration false and true.

src/components/response/ResponseBlock.tsx
```diff
@@ -672,57 +673,79 @@ export function ResponseBlock({
           // Check if this response is in the current location
           const isInCurrentLocation = responses.some((r) => r.id === response.id);
 
-          if (isInCurrentLocation) {
-            // Increment index for each response, unless it is a textOnly response
-            if (response.type !== 'textOnly') {
+          const isNonQuestion = response.type === 'textOnly' || response.type === 'divider' || ('divider' in response && Boolean(response.divider));
```
JackWilb, line 676: No response type or generated schema defines a `divider` property; the supported response-level field is `withDivider`, while an actual divider is identified by `type: divider`. This compatibility branch cannot be reached by a valid parsed config and adds unnecessary complexity. Please remove it.

JackWilb (changes requested):
Thanks for the focused delay and enumeration work. I found two existing behavior regressions that need to be addressed before merge: the added wrapper changes the response DOM structure used by shipped and external styles, and passing disabledAttempts into the inert wrapper overrides existing editability behavior in training and navigation flows. I also left smaller comments on the schema reference, delay validation, test coverage, and an unnecessary compatibility branch.

A7700:
delay until stimulus is satisfied

PR #1456, opened 2026-09-03
Deeply Insturmented Code Mirror Demo

mcnuttandrew:
Basically just a quick demo of a deeply insturmented code mirror


JackWilb (approved):
Thanks for the CodeMirror demo. The follow-up fixes and focused regression tests look good, and the current head's checks are green.

